Issue #4, opened 2018-06-21
Undefined variables in wrapper.py


wiegandm:
Both `scan_id` and `target` are undefined.

https://github.com/greenbone/ospd-openvas/blob/3f6586b538ee95dea0af403ce3018efa36e9fa73/ospd_openvas/wrapper.py#L207

PR #5, opened 2018-06-21
Fix issue #4.

jjnicola:
* ospd_openvas/wrapper.py (__init__): Log an error message and exit if
a redis connection is not possible.
* ospd_openvas/openvas_db.py (max_db_index): Fix exception error.

### DIFF
--- a/ospd_openvas/wrapper.py
+++ b/ospd_openvas/wrapper.py
@@ -22,10 +22,9 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA
 
-import logging
 import subprocess
 
-from ospd.ospd import OSPDaemon
+from ospd.ospd import OSPDaemon, logger
 from ospd.misc import main as daemon_main
 from ospd_openvas import __version__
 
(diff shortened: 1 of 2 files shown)


ospd_openvas/wrapper.py
```diff
@@ -202,10 +201,9 @@ def __init__(self, certfile, keyfile, cafile):
             self.add_scanner_param(name, param)
 
         if openvas_db.db_init() is False:
-            self.add_scan_error(
-                scan_id, host=target,
-                value='OpenVAS Redis Error: Not possible to find db_connection.')
-            return 2
+            logger.error('OpenVAS Redis Error: Not possible '
+                         'to find db_connection.')
+            exit()
```
wiegandm: I do not think exiting when `__init__` encounters a problem is a good solution. Wouldn't it be better to raise an exception?

wiegandm:
> Using add_scan_error() informs the user about problems in the scan.

Indeed, but the way `add_scan_error` was used would not have worked in an error case as it was using undefined parameters.

In this case the `OSPDaemon` instance fails to initialize and will not result in a viable scanner. I am not familiar enough with the API and the design to decide whether this should result in a `scan_error` or whether it should be handled in a different way.

jjnicola:
Yes, it is intentional. It is done during the scanner initialization, therefore I have no scan_id to add an scan_error to be retrieved by a client.

PR #6, opened 2018-06-28
Ospd-openvas is able to run a scan.

jjnicola:
Prepare the kb to run a scan.
Retrieve results from Redis.
When the scan finished, clean the redis kb.
If the scan is stopped with <stop_scan scan_id=...>, stop the scan cleanly and delete the redis kb. 
Add custom NVTs metadata to the <get_vts/> response.

### DIFF
--- a/ospd_openvas/nvticache.py
+++ b/ospd_openvas/nvticache.py
@@ -1,0 +1,112 @@
+# -*- coding: utf-8 -*-
+# Description:
+# Provide functions to handle NVT Info Cache
+#
+# Authors:
+# Juan José Nicola <juan.nicola@greenbone.net>
+#
+# Copyright:
+# Copyright (C) 2018 Greenbone Networks GmbH
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA 02110-1301 USA
+
+""" Functions related to the NVT information. """
+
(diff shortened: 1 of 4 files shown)


ospd_openvas/nvticache.py
```diff
@@ -0,0 +1,112 @@
...
+    resp = ctx.smembers('oid:%s:prefs' % oid)
+    timeout = ctx.lindex('nvt:%s' % oid,
+                         openvas_db.nvt_meta_fields.index("NVT_TIMEOUT_POS"))
+    preferences = ET.Element('preferences')
+    if int(timeout) > 0:
+        xml_timeout = ET.Element('timeout')
+        xml_timeout.text = timeout
+        preferences.append(xml_timeout)
+
+    if len(resp) > 0:
```
wiegandm: You could make use of the fact that empty sequences are false and simplify this to `if resp:` if you think it would improve readability, see https://www.python.org/dev/peps/pep-0008/#programming-recommendations

ospd_openvas/nvticache.py
```diff
@@ -0,0 +1,112 @@
...
+
+def get_nvt_all(oid, is_custom=0):
+    """ Get a full NVT. Returns an XML tree with the NVT metadata.
+    """
+    ctx = openvas_db.get_kb_context()
+
+    resp = ctx.lrange("nvt:%s" % oid,
+                      openvas_db.nvt_meta_fields.index("NVT_FILENAME_POS"),
+                      openvas_db.nvt_meta_fields.index("NVT_VERSION_POS"))
+    if (isinstance(resp, list) and len(resp) > 0) is False:
```
wiegandm: See above, `len(resp) > 0` could be simplified to `resp` if so desired -- or even reduce it to: `if not isinstance(resp, list):` since `isinstance(False, list)` is `False`.

ospd_openvas/wrapper.py
```diff
@@ -228,14 +248,29 @@ def load_vts(self):
...
             if ret == -1:
                 logger.info("Dupplicated VT with OID: {0}".format(vt_id))
             if ret == -2:
                 logger.info("{0}: Invalid OID.".format(vt_id))
 
+    def get_custom_vt_as_xml_str(self, custom):
+        """ Create a string representation of the XML object from the
+        custom data object.
+
+        The custom XML object which is returned will be embedded
```
wiegandm: Does the function actually do what its docstring describes? As far as I can tell it merely returns its parameter `custom` as-is.

Also, consider declaring this method a static method as it does not look like it modifies the instance.

PR #9, opened 2018-07-05
Pass vt_params to add_vt().

jjnicola:
Implement get_params_vt_as_xml_str().
Modify get_nvt_pref() to return in the right xml format, adding
attributes and subelements to the main element <vt_params>.


jjnicola:
Depends on greenbone/ospd#28

PR #10, opened 2018-07-11
Get the vts list as a dictionary with their params.

jjnicola:


janowagner:
Merging despite the failing test because that failure is due to the fact
that ospd is not available via pypi and does not find its way to the test setup in other ways either.
That needs to be fixed with a separate change.

PR #16, opened 2018-09-07
Performance improvements.

jjnicola:
Improve performance using select instead a new connection.
Improve performance getting results from a list of host Kbs instead of search them in all redis dbs.


jjnicola:
Depends on greenbone/openvas-scanner#164.

PR #17, opened 2018-09-21
Process nvt families passed in vtgroups.

jjnicola:
Add plugins to the list from a specific family passed by vtgroups.


jjnicola:
It depends on greenbone/ospd#45.

PR #21, opened 2018-10-15
Fix scan-stop.

jjnicola:
Let openvassd to do the redis clean up when the scan is stopped.


ospd_openvas/wrapper.py
```diff
@@ -524,6 +519,7 @@ def build_credentials_as_prefs(credentials):
 
         return cred_prefs_list
 
+
```
kroosec: Any reason for adding this empty line ?

PR #22, opened 2018-10-18
get_nvt_params() returns a dictionary instead of an XML element.

jjnicola:
Add get_nvt_prefs().
Add get_nvt_timeout().
get_params_vt_as_xml_str() create the XML element with parameters
for <get_vts> command.


jjnicola:
The Codacy issue about "Using xml.etree.ElementTree to parse untrusted XML data..." appears because it imports xml.etree.ElementTree for tostring(), Element() and SubElement() which don't have a replacement in defusedxml.

PR #23, opened 2018-10-19
Add get_vt_param_type() and check_param_type().

jjnicola:
Since the vt preference type is not part of the OSP protocol
in the vt_selection member of start_scan command, it takes the
type from the vts dictionary and checks if the value is consistent
with the type.


jjnicola:
Depends on PR greenbone/ospd#58 and greenbone/ospd-openvas#22

PR #25, opened 2018-10-24
Add vt references to the vts dictionary.

jjnicola:
Add get_nvt_refs(), which returns a dictionary with vt references.
Since the references are an element it self, don't add it in
metadata (custom element).

Add get_refs_vt_as_xml_str().
Return an XML element as string with the vt references.
It is the content of <vt_refs> of each vt in the response to <get_vts>.


jjnicola:
Depends on PR greenbone/ospd#60

PR #35, opened 2018-11-22
Automatic VT update

jjnicola:
A feed version check is done periodically. It will be updated only if there is no running scans. By the other hand, a new scan will not be run if there is a pending feed update, and a message error will be set as result.


bjoernricks:
Using global variables is often a sign of immature design of the software. So introducing more and more global statements should be avoided.

PR #36, opened 2018-11-27
Switch from unsorted set to ordered list.

jjnicola:
Depends on PR greenbone/gvm-libs#152.

### DIFF
--- a/ospd_openvas/nvticache.py
+++ b/ospd_openvas/nvticache.py
@@ -143,7 +143,8 @@ def get_nvt_refs(oid):
 
 def get_nvt_prefs(ctx, oid):
     """ Get NVT preferences. """
-    prefs = ctx.smembers('oid:%s:prefs' % oid)
+    key = ('oid:%s:prefs' % oid)
+    prefs = ctx.lrange(key, 0, -1)
     return prefs
 
 def get_nvt_timeout(ctx, oid):
--- a/ospd_openvas/openvas_db.py
+++ b/ospd_openvas/openvas_db.py
@@ -179,33 +179,33 @@ def get_kb_context():
 
     return REDISCONTEXT
 
-def item_get_set(name):
-    """ Get all values under a KB elements.
+def item_get_list(name):
+    """ Get all values under a KB key list.
     The right global REDISCONTEXT must be already set.
     """
     ctx = get_kb_context()
-    return ctx.smembers(name)
+    return ctx.lrange(name, 0, -1)
 
-def remove_set_member(key, member):
-    """ Remove member of the key set.
+def remove_list_item(key, value):
+    """ Remove item from the key list.
     The right global REDISCONTEXT must be already set.
     """
     ctx = get_kb_context()
-    ctx.srem(key, member)
+    ctx.lrem(key, 0, value)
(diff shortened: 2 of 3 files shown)


ospd_openvas/nvticache.py
```diff
@@ -143,7 +143,8 @@ def get_nvt_refs(oid):
 
 def get_nvt_prefs(ctx, oid):
     """ Get NVT preferences. """
-    prefs = ctx.smembers('oid:%s:prefs' % oid)
+    key = ('oid:%s:prefs' % oid)
```
bjoernricks: I suppose this isn't a python tuple so please remove the braces. If it should be a tuple please a a colon at the end.

ospd_openvas/nvticache.py
```diff
@@ -143,7 +143,8 @@ def get_nvt_refs(oid):
...
+    prefs = ctx.lrange(key, 0, -1)
```
bjoernricks: what's the meaning of 0 and -1 values. Either use keyword arguments (aka. kwargs) or module constants for this.

bjoernricks: Maybe even use both kwargs and constants here.

ospd_openvas/openvas_db.py
```diff
@@ -179,33 +179,33 @@ def get_kb_context():
...
 
-def item_get_set(name):
-    """ Get all values under a KB elements.
+def item_get_list(name):
+    """ Get all values under a KB key list.
     The right global REDISCONTEXT must be already set.
     """
     ctx = get_kb_context()
-    return ctx.smembers(name)
+    return ctx.lrange(name, 0, -1)
```
bjoernricks: Some as above. What's the meaning of 0 and -1?

ospd_openvas/openvas_db.py
```diff
@@ -179,33 +179,33 @@ def get_kb_context():
...
 
-def remove_set_member(key, member):
-    """ Remove member of the key set.
+def remove_list_item(key, value):
+    """ Remove item from the key list.
     The right global REDISCONTEXT must be already set.
     """
     ctx = get_kb_context()
-    ctx.srem(key, member)
+    ctx.lrem(key, 0, value)
```
bjoernricks: Same.

ospd_openvas/openvas_db.py
```diff
@@ -179,33 +179,33 @@ def get_kb_context():
...
 
 def item_get_single(name):
     """ Get a single KB element. The right global REDISCONTEXT must be
     already set.
     """
     ctx = get_kb_context()
-    return ctx.srandmember(name)
+    return ctx.lindex(name, 0)
```
bjoernricks: Same

ospd_openvas/openvas_db.py
```diff
@@ -232,7 +232,7 @@ def get_pattern(pattern):
 
     elem_list = []
     for item in items:
-        elem_list.append([item, ctx.smembers(item)])
+        elem_list.append([item, ctx.lrange(item, 0, -1)])
```
bjoernricks: same

PR #40, opened 2018-12-03
 Add cvss_base tag to the vts dictionary to fix result

jjnicola:
Also improve code style.


jjnicola:
The cvss_base value is only saved into the vts dictionary because we need it to insert it in the results. It is not send as a vt subelement, so it is not part of the protocol.
Other possibilities are:
- Insert the vector instead of the value in results.
- I can implement a function to calculate this value from the cvss_base_vector, to be inserted it in the result. I think this make sense if the cvss_base value will be removed soon from the tags.

janowagner:
The best option to me is to implement the calculation into the ospd base module as this
is of general interest and might be used by other wrappers as well.
Ideally already implement also the CVSSv3 calculation for later use.

jjnicola:
It depends on PR  greenbone/ospd#88.

It use the new CVSS class to calculate the severity score.

PR #46, opened 2018-12-20
Add unit test.

jjnicola:
Add unit tests for wrapper.py.
Add unit tests for nvticache.py
Add unit tests for db.py
Fix unit test for get_nvt_metadata.

PR #47, opened 2018-12-23
Update with new standard header.

janowagner:


wiegandm:
Tests are failing due to a unit tests not adapted to recent changes. As far as I can tell, the fix is in 48f3a2d35cba8584a8ebf38d179db0b23acb6b5c in #46.

PR #54, opened 2019-03-14
Add ID element to the NVT.

jjnicola:
Get the id element from redis and stored it in the dictionary.
Send the ID in the response of the <get_vts> command as vt_param argument.
e.g.
`<get_vts_response status="200" status_text="OK">
  <vts>
    <vt id="1.3.6.1.4.1.25623.1.0.12288">
      <name>Global variable settings</name>
      <vt_params>
        <vt_param id="6" type="checkbox">
          <name>Enable generic web application scanning</name>
          <default>no</default>
        </vt_param>
...`


jjnicola:
Depends on greenbone/openvas-scanner#282

PR #55, opened 2019-03-18
Use refs instead vt_refs.

jjnicola:
Use refs instead vt_refs.
Update tests accordingly.


jjnicola:
Depends on greenbone/ospd#93 .

PR #57, opened 2019-03-20
Add custom date formatter to be used by the nvt filter.

jjnicola:
Each wrapper has different formats for datetime values. This method
formats the datetime valued to be comapared with a standar format.

The method overwrites the original in the parent class.

### DIFF
--- a/ospd_openvas/wrapper.py
+++ b/ospd_openvas/wrapper.py
@@ -222,7 +222,7 @@ class OSPDopenvas(OSPDaemon):
 
     """ Class for ospd-openvas daemon. """
 
-    def __init__(self, certfile, keyfile, cafile, customvtfilter):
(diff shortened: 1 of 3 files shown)


jjnicola:
Depends on greenbone/ospd#94

ospd_openvas/wrapper.py
```diff
@@ -210,7 +226,8 @@ def __init__(self, certfile, keyfile, cafile):
         """ Initializes the ospd-openvas daemon's internal data. """
 
         super(OSPDopenvas, self).__init__(certfile=certfile, keyfile=keyfile,
```
bjoernricks: Unrelated to this PR: Could be simplified with python 3
```suggestion
        super().__init__(certfile=certfile, keyfile=keyfile,
```

PR #62, opened 2019-04-04
Don't crash if redis connection is lost or redis is flushed.

jjnicola:


ospd_openvas/db.py
```diff
@@ -153,15 +155,26 @@ def kb_connect(self, dbnum=0):
...
-        except ConnectionError as e:
+        tries = 5
+        while (tries):
+            try:
+                ctx = redis.Redis(unix_socket_path=self.db_address,
+                                  db=dbnum,
+                                  socket_timeout=SOCKET_TIMEOUT, encoding="latin-1",
+                                  decode_responses=True)
+                ctx.keys("test")
+            except:
```
bjoernricks: I am not 100 percent sure if a "catch all expections" is good here. Maybe only catch expected execptions for the connection errors.

jjnicola: I will improve it in a next PR. Thanks a lot!

PR #64, opened 2019-04-10
Changes related to resume tasks in ospd. 

jjnicola:


jjnicola:
Depends on greenbone/ospd#100

PR #66, opened 2019-04-15
Fix scan progress calculation, since it changed in ospd.

jjnicola:
Update test because target_legacy string was remove from scans table.

### DIFF
--- a/ospd_openvas/wrapper.py
+++ b/ospd_openvas/wrapper.py
@@ -699,7 +699,7 @@ class OSPDopenvas(OSPDaemon):
         return True
 
     def update_progress(self, scan_id, target, msg):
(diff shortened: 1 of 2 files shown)


jjnicola:
Depends on greenbone/ospd#101. Once merged and running the tests again it should pass without problems.

ospd_openvas/wrapper.py
```diff
@@ -699,26 +699,23 @@ def check(self):
         return True
 
     def update_progress(self, scan_id, target, msg):
-        """ Calculate percentage and update the scan status of a target
+        """ Calculate percentage and update the scan status of a hots
```
```suggestion
        """ Calculate percentage and update the scan status of a host
```

PR #69, opened 2019-05-17
 Fix host_details log for dead hosts and its scan progress.

jjnicola:


jjnicola:
@bjoernricks Thanks a lot!

PR #70, opened 2019-05-17
Update version before tagging.

jjnicola:
Update ospd requirement.


jjnicola:
Depends on PR greenbone/ospd#104. Once ospd has the version updated, circleci should build and test without problem.

PR #71, opened 2019-05-21
Update CHANGES before tagging.

jjnicola:
Quoting to @bjoernricks from PR #70:

> In general updating to a lower version (1.0b1 to 1.0a1) is not the best idea but IMHO it was a mistake to use a beta version in master and we don't have any real release yet. Therefore it should be ok.

### DIFF
--- a/CHANGES
+++ b/CHANGES
@@ -8,11 +8,8 @@ $ git clone https://github.com/greenbone/ospd-openvas.git
 $ cd ospd-openvas && git log
 
 
-ospd-openvas 1.0+beta1 (2018-XX-XX)
+ospd-openvas 1.0alpha1 (2019-05-21)
 
-This is the first beta release of the Open Scanner Protocol (OSP) server
+This is the first alpha version of the Open Scanner Protocol (OSP) server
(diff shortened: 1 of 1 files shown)


bjoernricks:
I would love if we could change the CHANGES file to CHANGELOG.md and also use the keepachangelog format like https://github.com/greenbone/gvm-tools/blob/master/CHANGELOG.md too.

wiegandm:
> I would love if we could change the CHANGES file to CHANGELOG.md and also use the keepachangelog format like https://github.com/greenbone/gvm-tools/blob/master/CHANGELOG.md too.

Indeed. I would also suggest a consistent version scheme, at least between the
Python modules. IIUC, this would mean the upcoming release should be named
`1.0.0.alpha1` instead of `1.0alpha1`.

bjoernricks:
> Indeed. I would also suggest a consistent version scheme, at least between the
Python modules. IIUC, this would mean the upcoming release should be named
1.0.0.alpha1 instead of 1.0alpha1.

I guess 1.0a1 would be the pep440 compliant version number. Can be checked with https://github.com/greenbone/gvm-tools/blob/master/gvmtools/__init__.py#L41

PR #78, opened 2019-06-03
Scan's niceness.

jjnicola:
Receive niceness argument.


jjnicola:
Depends on greenbone/ospd#109

PR #85, opened 2019-06-25
Fix stop scan cleanup

jjnicola:
As openvas is started with sudo, ospd-openvas (launched with non-root user permissions)
is not allowed to send SIGUSR2 kill signal to openvas and stop the running scan.
It used now instead, the new openvas command line option --stop-scan with sudo.
The new called openvas process will stop the running scan.

### DIFF
--- a/ospd_openvas/wrapper.py
+++ b/ospd_openvas/wrapper.py
@@ -913,10 +913,19 @@ class OSPDopenvas(OSPDaemon):
                     logger.debug(
                         'Process with pid %s already stopped', ovas_pid
                     )
-                self.openvas_db.release_db(current_kbi)
                 if parent:
-                    parent.send_signal(signal.SIGUSR2)
+                    cmd = ['sudo', 'openvas', '--scan-stop', scan_id]
+
+                    try:
+                        result = subprocess.Popen(cmd, shell=False)
+                    except OSError:
+                        # the command is not available
+                        logger.debug('Not possible to Stopping process: %s',
+                                     parent)
+                        return False
+
                     logger.debug('Stopping process: %s', parent)
+                self.openvas_db.release_db(current_kbi)
 
     def get_vts_in_groups(self, filters):
         """ Return a list of vts which match with the given filter.
--- a/ospd_openvas/wrapper.py
+++ b/ospd_openvas/wrapper.py
@@ -24,7 +24,6 @@
 import logging
 import subprocess
 import time
-import signal
 import uuid
 
 from os import path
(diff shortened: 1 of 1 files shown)


ospd_openvas/wrapper.py
```diff
@@ -913,10 +913,19 @@ def stop_scan_cleanup(
                     logger.debug(
                         'Process with pid %s already stopped', ovas_pid
                     )
-                self.openvas_db.release_db(current_kbi)
                 if parent:
-                    parent.send_signal(signal.SIGUSR2)
+                    cmd = ['sudo', 'openvas', '--scan-stop', scan_id]
+
+                    try:
+                        result = subprocess.Popen(cmd, shell=False)
```
bjoernricks: Any reason why not just
```python3
subprocess.run(cmd)
```
you don't even use the assigned result

ospd_openvas/wrapper.py
```diff
@@ -913,10 +913,19 @@ def stop_scan_cleanup(
...
+                    except OSError:
+                        # the command is not available
```
bjoernricks: the log message should also include the reason from the error

ospd_openvas/wrapper.py
```diff
@@ -913,10 +913,19 @@ def stop_scan_cleanup(
...
+                        logger.debug('Not possible to Stopping process: %s',
```
```suggestion
                        logger.debug('Not possible to Stopping process: %s. Reason %s',
```

ospd_openvas/wrapper.py
```diff
@@ -913,10 +913,19 @@ def stop_scan_cleanup(
...
                         'Process with pid %s already stopped', ovas_pid
                     )
-                self.openvas_db.release_db(current_kbi)
                 if parent:
-                    parent.send_signal(signal.SIGUSR2)
+                    cmd = ['sudo', 'openvas', '--scan-stop', scan_id]
+
+                    try:
+                        result = subprocess.Popen(cmd, shell=False)
+                    except OSError:
```
```suggestion
                    except OSError as e:
```

ospd_openvas/wrapper.py
```diff
@@ -913,10 +913,19 @@ def stop_scan_cleanup(
...
+                        # the command is not available
+                        logger.debug('Not possible to Stopping process: %s',
+                                     parent)
```
```suggestion
                                     parent, e)
```

ospd_openvas/wrapper.py
cfi-gb: It could also make sense to include the requirement for a full / correct sudo setup in the installation docs.

jjnicola: In future, if sudo fails, it will call openvas without sudo.

cfi-gb: Which probably won't solve the issue if e.g. openvas is started as root (but without sudo) and ospd-openvas is started as non-root user.

PR #87, opened 2019-06-25
Check sudo permissions before call openvas.

jjnicola:
Check if the user running ospd-openvas has permissions to call openvas
with sudo and without password.
Depending on it, it will run openvas with sudo or without it.

How to give the user permission to call openvas with sudo and without password:
`$ sudo visudo
`
Add at the end of the file
`<user> ALL = NOPASSWD: /<install/path/sbin/openvas
`

### DIFF
--- a/ospd_openvas/wrapper.py
+++ b/ospd_openvas/wrapper.py
@@ -269,7 +269,7 @@ class OSPDopenvas(OSPDaemon):
         self.scanner_info['description'] = OSPD_DESC
         for name, param in OSPD_PARAMS.items():
             self.add_scanner_param(name, param)
-        self.sudo_available = self.sudo_check()
+        self.sudo_available = False
 
         self.scan_only_params = dict()
         self.main_kbindex = None
@@ -718,7 +718,6 @@ class OSPDopenvas(OSPDaemon):
 
     def sudo_check(self):
         """ Checks that sudo is available and set the global var. """
-        _sudo_available = False
         try:
             result = subprocess.check_call(
                 ['sudo', '-n', 'openvas', '-s'], stdout=subprocess.PIPE
@@ -726,12 +725,10 @@ class OSPDopenvas(OSPDaemon):
         except subprocess.CalledProcessError as e:
             logger.debug('It was not possible to call openvas with sudo. '
                          'The scanner will run as non-root user. Reason %s', e)
-            return _sudo_available
+            self.sudo_available = False
 
         if result == 0:
-            _sudo_available = True
-
-        return _sudo_available
+            self.sudo_available = True
 
     def check(self):
         """ Checks that openvas command line tool is found and
--- a/ospd_openvas/wrapper.py
+++ b/ospd_openvas/wrapper.py
@@ -928,10 +928,9 @@ class OSPDopenvas(OSPDaemon):
                         'Process with pid %s already stopped', ovas_pid
                     )
                 if parent:
+                    cmd = ['openvas', '--scan-stop', scan_id]
                     if self.sudo_available:
-                        cmd = ['sudo', 'openvas', '--scan-stop', scan_id]
-                    else:
-                        cmd = ['openvas', '--scan-stop', scan_id]
+                        cmd = ['sudo'] + cmd
 
                     try:
                         subprocess.Popen(cmd, shell=False)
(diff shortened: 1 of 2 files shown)


ospd_openvas/wrapper.py
```diff
@@ -715,6 +716,23 @@ def get_detection_vt_as_xml_str(
...
+        """ Checks that sudo is available and set the global var. """
+        _sudo_available = False
+        try:
+            result = subprocess.check_call(
+                ['sudo', '-n', 'openvas', '-s'], stdout=subprocess.PIPE
+            )
+        except subprocess.CalledProcessError as e:
+            logger.debug('It was not possible to call openvas with sudo. '
+                         'The scanner will run as non-root user. Reason %s', e)
+            return _sudo_available
```
```suggestion
            self._sudo_available = False
```

ospd_openvas/wrapper.py
```diff
@@ -715,6 +716,23 @@ def get_detection_vt_as_xml_str(
 
         return tostring(_detection).decode('utf-8')
 
+    def sudo_check(self):
+        """ Checks that sudo is available and set the global var. """
+        _sudo_available = False
+        try:
+            result = subprocess.check_call(
+                ['sudo', '-n', 'openvas', '-s'], stdout=subprocess.PIPE
+            )
```
```suggestion
            )
            self._sudo_available = True
```

ospd_openvas/wrapper.py
```diff
@@ -715,6 +716,23 @@ def get_detection_vt_as_xml_str(
...
+        except subprocess.CalledProcessError as e:
+            logger.debug('It was not possible to call openvas with sudo. '
+                         'The scanner will run as non-root user. Reason %s', e)
+            return _sudo_available
+
+        if result == 0:
+            _sudo_available = True
+
+        return _sudo_available
```
bjoernricks: Just remove these lines above.

ospd_openvas/wrapper.py
```diff
@@ -715,6 +716,23 @@ def get_detection_vt_as_xml_str(
 
         return tostring(_detection).decode('utf-8')
 
+    def sudo_check(self):
+        """ Checks that sudo is available and set the global var. """
+        _sudo_available = False
+        try:
+            result = subprocess.check_call(
```
bjoernricks: check_call doesn't return the returncode. If the returncode is != 0 it raises an exception https://docs.python.org/3/library/subprocess.html#subprocess.check_output So no need to check the result.

ospd_openvas/wrapper.py
```diff
@@ -715,6 +716,23 @@ def get_detection_vt_as_xml_str(
 
         return tostring(_detection).decode('utf-8')
 
+    def sudo_check(self):
```
```suggestion
    @property
    def sudo_available(self):
```

ospd_openvas/wrapper.py
```diff
@@ -269,6 +269,7 @@ def __init__(self, certfile, keyfile, cafile, niceness=None):
         self.scanner_info['description'] = OSPD_DESC
         for name, param in OSPD_PARAMS.items():
             self.add_scanner_param(name, param)
+        self.sudo_available = self.sudo_check()
```
```suggestion
        self._sudo_available = None

File: ospd_openvas/wrapper.py [hunk 7]
```diff
@@ -715,6 +716,23 @@ def get_detection_vt_as_xml_str(
 
         return tostring(_detection).decode('utf-8')
 
+    def sudo_check(self):
+        """ Checks that sudo is available and set the global var. """
```
```suggestion
        """ Checks that sudo is available and set the global var. """
        if self._sudo_available is not None:
            return self._sudo_available
```

ospd_openvas/wrapper.py
```diff
@@ -715,6 +716,23 @@ def get_detection_vt_as_xml_str(
...
+        _sudo_available = False
```
```suggestion
```

ospd_openvas/wrapper.py
```suggestion
            subprocess.check_call(
```

ospd_openvas/wrapper.py
```diff
@@ -913,7 +931,10 @@ def stop_scan_cleanup(
                         'Process with pid %s already stopped', ovas_pid
                     )
                 if parent:
-                    cmd = ['sudo', 'openvas', '--scan-stop', scan_id]
+                    if self.sudo_available:
+                        cmd = ['sudo', 'openvas', '--scan-stop', scan_id]
+                    else:
+                        cmd = ['openvas', '--scan-stop', scan_id]
```
bjoernricks: These lines can be simplified by something like:
```python3
cmd = ['openvas', '--scan-stop', scan_id]
if self.sudo_available:
    cmd = ['sudo'] + cmd

File: ospd_openvas/wrapper.py [hunk 10]
```diff
@@ -1264,7 +1285,11 @@ def exec_scan(self, scan_id, target):
             value='An OpenVAS Scanner was started for %s.' % target,
         )
 
-        cmd = ['sudo', 'openvas', '--scan-start', openvas_scan_id]
+        if self.sudo_available:
+            cmd = ['sudo', 'openvas', '--scan-start', openvas_scan_id]
+        else:
+            cmd = ['openvas', '--scan-start', openvas_scan_id]
+
```
bjoernricks: Can be simplified as above.

Maybe it would even better to extract calling openvas into an own method/function e.g.
```python3
def call_openvas(args, use_sudo=False, niceness=None):
   cmd = args

   if use_sudo:
       cmd = ['sudo'] + cmd

   if niceness is not None:
      cmd = ['nice', ...] + cmd

    try:
        subproces. ....
    ...
```
but i am not 100 percent sure about that because we are calling different functions of subrocess at the moment with slightly different arguments.

PR #89, opened 2019-07-05
Expect the parameter id of the VT instead of the name.

jjnicola:
Also fix check_type, since it has expected yes/no instead of 1/0.

### DIFF
--- a/ospd_openvas/wrapper.py
+++ b/ospd_openvas/wrapper.py
@@ -520,10 +520,10 @@ class OSPDopenvas(OSPDaemon):
             string: xml element as string.
         """
         vt_params_xml = Element('params')
-        for _pref_name, prefs in vt_params.items():
+        for _pref_id, prefs in vt_params.items():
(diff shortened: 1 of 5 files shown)


ospd_openvas/wrapper.py
```diff
@@ -971,11 +971,20 @@ def get_vts_in_groups(self, filters):
...
             return vt_params_list[vt_param_id]["type"]
         return False
 
+    def get_vt_param_name(self, vtid, vt_param_id):
+        """ Return the type of the vt parameter from the vts dictionary. """
+
+        vt_params_list = self.vts[vtid].get("vt_params")
+        if vt_params_list.get(vt_param_id):
+            return vt_params_list[vt_param_id]["name"]
+        return False
```
bjoernricks: I am not sure about the return type. Shouldn't it be None if the param isn't available?

jjnicola:
Depends on PR greenbone/gvmd#636

PR #103, opened 2019-08-07
Openvas config files

bjoernricks:
Add config files to allow running ospd-openvas with redis-server from the system

Howto use:
```sh
cp openvas.conf <install prefix>/etc/openvas/
sudo cp redis-openvas.conf /etc/redis/
sudo chown redis:redis /etc/redis/redis-openvas.conf
sudo systemctl start redis-server@openvas
```

### DIFF
--- a/openvas.conf
+++ b/openvas.conf
@@ -1,0 +1,1 @@
+db_address = /var/run/redis-openvas/redis.sock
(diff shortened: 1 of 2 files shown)


bjoernricks:
Withdrawn as discussed with @jjnicola

PR #105, opened 2019-08-16
Fix feed outdated

jjnicola:
This prevent a scan to be started if the feed is not loaded.


ospd_openvas/daemon.py
```diff
@@ -26,8 +26,7 @@
 import time
 import uuid
 
-from os import path
-
+from pathlib import Path
```
```suggestion
from pathlib import Path

```

PR #112, opened 2019-09-19
Change parsing of modification time

janowagner:
With https://github.com/greenbone/gvm-libs/pull/265 the timestamps
in redis created by openvas will be always seconds since epoch as
strings.

This patch changes the parsing of the string to assuming the string
is seconds since epoch.

### DIFF
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -26,12 +26,12 @@ import subprocess
 import time
 import uuid
 
+from datetime import datetime
+
 from pathlib import Path
 from lxml.etree import tostring, SubElement, Element
 
 import psutil
-
-from datetime import datetime
 
 from ospd.errors import OspdError
 from ospd.ospd import OSPDaemon
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -235,15 +235,12 @@ class OpenVasVtsFilter(VtsFilter):
     """
 
     def format_vt_modification_time(self, value):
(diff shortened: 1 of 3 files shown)


ospd_openvas/daemon.py
```diff
@@ -235,15 +235,12 @@ class OpenVasVtsFilter(VtsFilter):
     """
 
     def format_vt_modification_time(self, value):
-        """ Convert the datetime value in an 19 character string
-        representing YearMonthDateHourMinuteSecond.
+        """ Convert the string seconds since epoch into a 19 character
+        string representing YearMonthDayHourMinuteSecond.
```
bjoernricks: It doesn't state anything about the timezone of the returned string. We should make it clear in which timezone this string representation is in. I would prefer UTC.

ospd_openvas/daemon.py
```diff
@@ -235,15 +235,12 @@ class OpenVasVtsFilter(VtsFilter):
...
         e.g. 20190319122532
         """
 
-        date = value[7:26].replace(" ", "")
-        date = date.replace("-", "")
-        date = date.replace(":", "")
-        return date
+        return time.strftime("%Y%m%d%H%M%S", time.gmtime(int(value)))
```
bjoernricks: I would prefere fromtimestamp here and at best utcfromtimestamp https://docs.python.org/3/library/datetime.html#datetime.datetime.utcfromtimestamp
```suggestion
        return datetime.utcfromtimestamp(int(value)).strftime("%Y%m%d%H%M%S")
```

PR #117, opened 2019-09-27
Check if it is running as root.

jjnicola:
This avoid to call openvas with sudo when the daemon is already running as root user.


bjoernricks:
Just a linting error needs to be adjusted.

PR #130, opened 2019-10-25
[1.0] Check the vt's preference value for type 'file'.

jjnicola:
If the preference type is file, it check if the value is
encoded into base64 format, as expected for openvas.

### DIFF
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -25,8 +25,10 @@ import logging
 import subprocess
 import time
 import uuid
+import binascii
 
 from datetime import datetime
+from base64 import b64decode
 
 from pathlib import Path
 from os import geteuid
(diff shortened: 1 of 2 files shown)


ospd_openvas/daemon.py
```diff
@@ -25,13 +25,18 @@
...
+import binascii
 
 from datetime import datetime
+from base64 import b64decode
 
 from pathlib import Path
 from os import geteuid
 from lxml.etree import tostring, SubElement, Element
 
+from base64 import b64decode
```
```suggestion
```

ospd_openvas/daemon.py
```diff
@@ -25,13 +25,18 @@
...
+import binascii
```
```suggestion
```

ospd_openvas/daemon.py
```diff
@@ -25,13 +25,18 @@
...
+
```
```suggestion
```

Issue #125, opened 2019-10-21
[1.0.0] Error while connecting to redis is not catched.


tuxmaster5000:
When the daemon can't read/write to the redis socket, then an python exception is thrown, but not catched.

>  File "/usr/bin/ospd-openvas", line 11, in <module>
>                                                                               load_entry_point('ospd-openvas==1.0.0', 'console_scripts', 'ospd-openvas')()
>                                                                             File "/usr/lib/python3.6/site-packages/ospd_openvas/daemon.py", line 1454, in main
>                                                                               daemon_main('OSPD - openvas', OSPDopenvas)
>                                                                             File "/usr/lib/python3.6/site-packages/ospd/main.py", line 159, in main
>                                                                               daemon.init()
>                                                                             File "/usr/lib/python3.6/site-packages/ospd_openvas/daemon.py", line 283, in init
>                                                                               self.openvas_db.db_init()
>                                                                             File "/usr/lib/python3.6/site-packages/ospd_openvas/db.py", line 139, in db_init
>                                                                               self.max_db_index()
>                                                                             File "/usr/lib/python3.6/site-packages/ospd_openvas/db.py", line 117, in max_db_index
>                                                                               ctx = self.kb_connect()
>                                                                             File "/usr/lib/python3.6/site-packages/ospd_openvas/db.py", line 195, in kb_connect
>                                                                               'Redis Error: Not possible to connect to the kb.'
>                                                                           ospd_openvas.errors.OspdOpenvasError: Redis Error: Not possible to connect to the kb.
>

PR #133, opened 2019-10-28
[1.0] Exit with exit code 1 if it was not possible to connect to redis.

jjnicola:
After trying 5 times to connect to redis, it exits with error code 1
instead of raise an error not being catched.


jjnicola:
Fix issue #125

PR #138, opened 2019-10-30
Improve error handling when creating vts xml elements.

jjnicola:
Avoid ospd to crash if there is an invalid char.


cfi-gb:
Fixes: https://github.com/greenbone/ospd/issues/169

PR #139, opened 2019-10-30
[1.0] Improve error handling when creating vts xml elements.

jjnicola:
Avoid ospd to crash if there is an invalid char.

### DIFF
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -626,7 +626,7 @@ class OSPDopenvas(OSPDaemon):
             _vt_dep = Element('dependency')
             try:
                 _vt_dep.set('vt_id', dep)
-            except TypeError:
+            except (ValueError, TypeError):
                 logger.error(
                     'Not possible to add dependency %s for vt %s', dep, vt_id
                 )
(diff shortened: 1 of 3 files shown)


ospd_openvas/daemon.py
```diff
@@ -1017,11 +1082,9 @@ def stop_scan_cleanup(
                 self.openvas_db.release_db(current_kbi)
                 for host_kb in range(0, self.openvas_db.max_dbindex):
                     self.openvas_db.select_kb(
-                        ctx,
```
mattmundell: These formatting changes would have been better in a separate commit.

ospd_openvas/daemon.py
```diff
@@ -626,7 +626,7 @@ def get_dependencies_vt_as_xml_str(
             _vt_dep = Element('dependency')
             try:
                 _vt_dep.set('vt_id', dep)
-            except TypeError:
```
mattmundell: This doesn't seem to be about improving the tests.

ospd_openvas/daemon.py
jjnicola: I did not realize about this since it is done automatically when I commit. There is a hook which runs black.

ospd_openvas/daemon.py
jjnicola: You are right. This improve the error handling, not the test. But it is related to the first commit. I should have done this with the first commit or in a separated one.

PR #148, opened 2019-11-08
Fix tests

jjnicola:


.pylintrc
```diff
@@ -47,7 +47,8 @@ confidence=
 disable=len-as-condition,
         attribute-defined-outside-init,
         missing-docstring,
-        bad-continuation
+        bad-continuation,
+        arguments-differ
```
bjoernricks: Should be replaced by an inline comment

See http://pylint.pycqa.org/en/latest/user_guide/message-control.html#block-disables

jjnicola:
@bjoernricks Thanks!

PR #150, opened 2019-11-12
Add set_nvticache_str().

jjnicola:
The nvticache name in redis (dependent on the gvm-libs) will be checked
for compatibility against a list of supported nvticache versions.

### DIFF
--- a/ospd_openvas/nvticache.py
+++ b/ospd_openvas/nvticache.py
@@ -23,7 +23,7 @@ import logging
 import subprocess
 import sys
 
-from distutils.version import StrictVersion
+from pkg_resources import parse_version
 
 from ospd_openvas.db import NVT_META_FIELDS
 from ospd_openvas.errors import OspdOpenvasError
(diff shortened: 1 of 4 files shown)


ospd_openvas/nvticache.py
```diff
@@ -20,14 +20,22 @@
...
 from ospd_openvas.db import NVT_META_FIELDS
+from ospd_openvas.errors import OspdOpenvasError
+
 
 logger = logging.getLogger(__name__)
 
 LIST_FIRST_POS = 0
 LIST_LAST_POS = -1
 
+SUPPORTED_NVTICACHE_VERSION = ['20.4']
```
bjoernricks: It should be named SUPPORTED_NVTICACHE_VERSION**S** and be a tuple to avoid changing the list. Tuples are immutable.

```suggestion
SUPPORTED_NVTICACHE_VERSIONS = ('20.4',)
```

PR #153, opened 2019-11-13
fix for work on redis 4.0.9

bartigor:
Corrected to work with redis 4.0.9 Maybe someone will come in handy

### DIFF
--- a/ospd_openvas/db.py
+++ b/ospd_openvas/db.py
@@ -287,7 +287,7 @@ class OpenvasDB(object):
 
         if not ctx:
             ctx = self.get_kb_context()
-        ctx.lrem(key, count=LIST_ALL, value=value)
+        ctx.lrem(key, LIST_ALL, value)
 
     def get_single_item(self, name, ctx=None, index=LIST_FIRST_POS):
         """ Get a single KB element.


bjoernricks:
I don't get why this should change anything. Instead of passing the necessary arguments as keyword they are now passed as positionals.

Beyond that I even don't know how this should affect redis server directly if we only call the redis python api.

bartigor:
With named arguments, this just doesn't work. I replaced https://redis-py.readthedocs.io/en/latest/#redis.Redis.lrem with positional ones according to the documentation and it worked.

bjoernricks:
I don't get that
```python
def foo(a, b, c):
   print(a, b, c)

>>> foo(1, 2, 3)
(1, 2, 3)
>>> foo(1, b=2, c=3)
(1, 2, 3)
```

bjoernricks:
Passing arguments via keywords is a standard python feature. This has nothing to do with the used redis library.

Even this works
```python
>>> foo(1, c=3, b=2)
(1, 2, 3)
```

bartigor:
I know that python should work this way and that, but why am I getting an error
lrem() got an unexpected keyword argument 'count'
Perhaps this is due to the implementation of the python-redis library.

bartigor:
I understand it seems there are different classes 
in StrictRedis(object)
```
  lrem(self, name, count, value)
```
in class Redis(StrictRedis):
```
def lrem(self, name, value, num=0):
```
and when you call class Redis(StrictRedis): it falls

PR #162, opened 2019-11-21
More Typing.

jjnicola:
- Add typing to db.py
- Improve typing in nvticache.py


jjnicola:
> Looks very good. I am not sure about some Optional returns because the haven't been mentioned in the docstrings but I guess you have checked them :-)

The returns were indeed checked. If a key is empty or it does not exists, redis return None. Therefore someones are optional.

PR #166, opened 2019-11-29
Improve nvti cache version handling

bjoernricks:


bjoernricks:
Needs to be backported to ospd-openvas 1.0 branch.

PR #167, opened 2019-11-29
Nvti cache updates

bjoernricks:
Backports of #165 and #166


jjnicola:
I tested and it works like a charm.

PR #171, opened 2019-12-03
Send messages generated by the scanner main process.

jjnicola:
Also remove trailing whitespaces from the host and OID strings.
Update test accordingly


ospd_openvas/daemon.py
```diff
@@ -1507,6 +1507,7 @@ def exec_scan(self, scan_id, target):
 
             ctx = self.openvas_db.kb_connect(self.main_kbindex)
             self.openvas_db.set_redisctx(ctx)
+            self.get_openvas_result(scan_id, "")
```
bjoernricks: The method name is confusing. A better name would be something like `collect_results_from_openvas`.

jjnicola: I will do it.

jjnicola:
@bjoernricks , Thanks for the reviews!

PR #182, opened 2020-01-17
Add support for alive test settings.

jjnicola:
Depends on PR greenbone/ospd#194. After merging ospd pr 194, the Pipefile.lock must be updated.

### DIFF
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -1195,10 +1195,10 @@ class OSPDopenvas(OSPDaemon):
 
         return 1
 
-    def process_vts(self, vts: List) -> Tuple[list, list]:
+    def process_vts(self, vts: List) -> Tuple[list, dict]:
         """ Add single VTs and their parameters. """
         vts_list = []
-        vts_params = []
+        vts_params = {}
         vtgroups = vts.pop('vt_groups')
 
         if vtgroups:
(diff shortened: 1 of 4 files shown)


ospd_openvas/daemon.py
```diff
@@ -229,6 +230,17 @@
...
+
+
+class AliveTest(IntEnum):
+    """ Alive Tests. """
+
+    ALIVE_TEST_TCP_ACK_SERVICE = (1,)
+    ALIVE_TEST_ICMP = (2,)
+    ALIVE_TEST_ARP = (4,)
+    ALIVE_TEST_CONSIDER_ALIVE = (8,)
+    ALIVE_TEST_TCP_SYN_SERVICE = 16
```
bjoernricks: Why are the all values besides ALIVE_TEST_TCP_SYN_SERVICE tuples and ALIVE_TEST_TCP_SYN_SERVICE not?

ospd_openvas/daemon.py
```diff
@@ -1362,6 +1373,109 @@ def build_credentials_as_prefs(credentials: Dict) -> List[str]:
...
+                added to the redis KB.
+        """
+        target_opt_prefs_list = []
+        if target_options and target_options.get('alive_test'):
+            try:
+                alive_test = int(target_options.get('alive_test'))
+            except ValueError:
+                logger.debug(
+                    'Alive test settings not applied. '
+                    'Invalid alive test value'
```
bjoernricks: The logger should print the invalid value

ospd_openvas/daemon.py
jjnicola: Not sure what happened here. Maybe autohooks? They should be all integer. Fixed.

PR #183, opened 2020-01-22
Check for a valid OID when prepare results.

jjnicola:
This improve the error handling and avoid unexpected traceback.


jjnicola:
> Just for curiosity when or why can this happen?

It was reported with the backtrace. I was not able to reproduce it. I guess it was a scan during a non finished feed update, when the VT was already present in the redis cache, but still not in the ospd vts dictionary.

PR #192, opened 2020-02-09
Create extended severities object

janowagner:
This is about the given values we might have at hand.
It is not yet considering multiple severities from a NVT though.
But if the internal severity structure has a origin or date,
then it will be considered already.


timopollmeier (changes requested):
The change itself looks okay, but the expected test output for `test_get_severities_xml` of `tests/test_daemon.py` needs to be adjusted.

timopollmeier (approved):

PR #194, opened 2020-02-14
Calculate vts sha256 hash after loading the vts in memory

jjnicola:
Depends on PR greenbone/ospd#232


jjnicola:
After PR greenbone/ospd#232 is merged, Pipfile.lock file must be updated.

PR #196, opened 2020-02-18
Start server before initialize the vts.

jjnicola:
Depends on PR greenbone/ospd#209


jjnicola:
> Looks good to me already! Is this PR for fixing the nasty could not connect to the scanner warnings in gvmd?

Yes. Now the socket is created during the init and it is able to respond to client with an error until it finishes the initialization.

PR #205, opened 2020-03-02
Get vts metadata from redis and reduce stored data in cache.

jjnicola:
Depends on PR greenbone/ospd#216. After merging, Pipfile.lock must be updated.

### DIFF
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -447,10 +447,6 @@ class OSPDopenvas(OSPDaemon):
         _vt_refs = self.nvti.get_nvt_refs(vt_id)
         _custom = self.nvti.get_nvt_metadata(vt_id)
 
-        if _vt_params is None or _custom is None:
-            logger.warning('Error loading VTs in memory. Trying again...')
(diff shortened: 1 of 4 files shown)


ospd_openvas/daemon.py
```diff
@@ -442,6 +442,116 @@ def scheduler(self):
...
+
+        return vt
+
+    def get_vt_iterator(self):
+        """ Yield the vts from the Redis NVTicache. """
+        oids = dict(self.nvti.get_oids())
+        for filename, vt_id in oids.items():
+            vt = self.get_single_vt(vt_id, oids)
+            yield (vt_id, vt)
+
```
bjoernricks: For me this is not something the daemon class should care off. Personally I would move this functionality in the NVTICache class.

jjnicola: Yes, sure. I am thinking on moving other methods to another files. But I would prefer to do the cleanup in a separated PR.

bjoernricks: Fine :-)

PR #206, opened 2020-03-04
Use lock file during feed update to avoid corrupted cache.

jjnicola:
Depends on PR greenbone/ospd#217


ospd_openvas/daemon.py
```diff
@@ -355,6 +361,43 @@ def feed_is_outdated(self, current_feed):
...
+
+    def create_feed_lock_file(self):
+        """ Create a lock file.
+            Return: True in success, False otherwise.
+        """
+        if self.feed_locked():
+            return False
+        else:
+            try:
+                with open(str(self.feed_lock_file), 'w') as f:
```
```suggestion
                with self.feed_lock_file.open('w') as f:
```
Details at https://docs.python.org/3/library/pathlib.html#pathlib.Path.open

ospd_openvas/daemon.py
```diff
@@ -259,6 +259,9 @@ def __init__(self, *, niceness=None, **kwargs):
 
         self._niceness = str(niceness)
 
+        self.feed_lock_file = (
+            Path(kwargs.get('lock_file_dir')) / 'feed-update.lock'
```
bjoernricks: Personally I wouldn't use the kwargs dict. Instead i would add `lock_file_dir='some/default/file/path/'` as an argument to `__init__`

PR #215, opened 2020-03-10
Add details parameter to get_vt_iterator()

jjnicola:
If details is True, it adds the vt dependencies as subelement
with the oid. For this an expensive query to redis is necessary.
The query takes about 3 seconds, which is especially expensive when
only one vt must be sent.
Otherwise, the dependencies are added as subelement of custom, the query
will not be perfomed, and quicker response is given for a single vt.

Depends on PR greenbone/ospd#222


jjnicola:
It needs a Pipfile.lock update, after greenbone/ospd#222 is merged

PR #223, opened 2020-03-17
Move scan prefs setup to a class

jjnicola:


ospd_openvas/daemon.py
```diff
@@ -1558,201 +1178,72 @@ def exec_scan(self, scan_id: str):
...
                 ),
             )
             return 2
 
-        ports = self.get_scan_ports(scan_id)
+        do_not_launch = False
+        kbdb = self.main_db.get_new_kb_database()
+        scan_prefs = PreferenceHandler(scan_id, kbdb, self.scan_collection)
+        scan_prefs.set_target()
+        ports = scan_prefs.set_ports()
```
bjoernricks: A set function/method should *usually* not return a value and expect some parameters.

ospd_openvas/preferencehandler.py
```diff
@@ -0,0 +1,610 @@
...
+        if self._target_options is not None:
+            return self._target_options
+
+        self._target_options = self.scan_collection.get_target_options(
+            self.scan_id
+        )
+        return self._target_options
+
+    def get_vts_in_groups(
+        self, filters: List[str], vts_cache: Dict[str, Dict],
```
bjoernricks: Personally i would move the vts_cache argument to the front for all get_vt(s)_ methods but it's just a mater of taste.

ospd_openvas/preferencehandler.py
```diff
@@ -0,0 +1,610 @@
...
+        credentials = self.scan_collection.get_credentials(self.scan_id)
+        if credentials:
+            cred_prefs = self.build_credentials_as_prefs(credentials)
+            if cred_prefs:
+                self.kbdb.add_scan_preferences(self.openvas_scan_id, cred_prefs)
+                return True
+
+        return False
+
+    def set_main_kbindex(self, main_kbindex: int):
```
bjoernricks: It the main_kbindex really necessary? I think it is the same as self.kbdb.index.

ospd_openvas/preferencehandler.py
```diff
@@ -0,0 +1,610 @@
...
+            families[family].append(oid)
+
+        for elem in filters:
+            key, value = elem.split('=')
+            if key == 'family' and value in families:
+                vts_list.extend(families[value])
+
+        return vts_list
+
+    def get_vt_param_type(
```
bjoernricks: Oh btw. are the get_vt(s)_ methods used outside of this class? if not they should be marked as internal by using an underscore prefix in their name `def _get_vt_`

ospd_openvas/preferencehandler.py
```diff
@@ -0,0 +1,610 @@
...
+    @property
+    def openvas_scan_id(self) -> str:
+        """ Create the openvas scan id and store it in the redis kb.
+        Return the openvas scan_id.
+        """
+        if self._openvas_scan_id is not None:
+            return self._openvas_scan_id
+
+        self._openvas_scan_id = str(uuid.uuid4())
+        self.kbdb.add_scan_id(self.scan_id, self._openvas_scan_id)
```
bjoernricks: Usually getters and properties should not have side effects. But I am unsure how to propagate the scan id to the kb if it wouldn't be done here.

ospd_openvas/preferencehandler.py
```diff
@@ -0,0 +1,610 @@
...
+
+                vts_params[
+                    "{0}:{1}:{2}:{3}".format(
+                        vtid, vt_param_id, param_type, param_name
+                    )
+                ] = str(vt_param_value)
+
+        return vts_list, vts_params
+
+    def set_plugins(self, vts_cache: Dict[str, Dict],) -> bool:
```
bjoernricks: Just a thought that comes to my mind may vts_cache change during the lifetime of the PreferenceHandler class? If not it could be a passed to the constructur instead of passing it to several methods.

ospd_openvas/preferencehandler.py
```diff
@@ -0,0 +1,610 @@
...
+            item = None
+            plugin_list = None
+            nvts = None
+
+            return True
+
+        return False
+
+    @staticmethod
+    def build_alive_test_opt_as_prefs(target_options: Dict) -> List[str]:
```
bjoernricks: Would be nice to get a more concrete type of the Dict here.

ospd_openvas/preferencehandler.py
```diff
@@ -0,0 +1,610 @@
...
+        """ Parse the target options dictionary.
+        @param credentials: Dictionary with the target options.
+
+        @return A list with the target options in string format to be
+                added to the redis KB.
```
bjoernricks: We should always use the [google style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) for the docstrings.
```suggestion
        Arguments:
            credentials: Dictionary with the target options.

        Return:
            A list with the target options in string format to be
                added to the redis KB.
```

ospd_openvas/preferencehandler.py
jjnicola: Yes, we need it on the OpenVAS side.

bjoernricks: I wanted to express if passing the main_kbindex is necessary here. imho we already have the information in self.kbdb. I think the passed main_kbindex is the same as self.kbdb here.

PR #226, opened 2020-03-27
handle dead hosts

ArnoStiefvater:
When the test_alive_hosts_only feature of openvas is enabled we
do not scan every single host in the target list and therefore
do not have the host progress for all hosts. To still be able
to update the progress bar we need to send all the dead hosts
from openvas to ospd-openvas. This is not an optimal solution as
we get big jumps in the progress bar if the target lists has many
hosts which are dead because they all get sent at once after the
host discovery is completed.


jjnicola:
close in favor of PR #229

PR #229, opened 2020-03-31
Write results and update host stauts in batch

jjnicola:
This allows to access the scan table less time. As the scan table is in a separated process, accessing the shared memory makes the writing of result for large scan quite slow.
Now this is improved.
This change is especially important since the new test_alive_host_only is implemented in openvas.
Large networks can be scanned and many host can be dead, generating a big amount of "host_details" results for the dead hosts. The dead host list  arrives to the ospd-openvas in list of up to 1000 hosts, which will be inserted and updated at once in the scan table.

Depends on PR greenbone/ospd#242

### DIFF
--- a/Pipfile.lock
+++ b/Pipfile.lock
@@ -72,29 +72,27 @@
         },
         "cryptography": {
             "hashes": [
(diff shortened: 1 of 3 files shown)


jjnicola:
Pipfile.lock update requiered.

PR #230, opened 2020-04-01
Add method to check if openvas process is still alive.

jjnicola:


ospd_openvas/daemon.py
```diff
@@ -1116,6 +1116,38 @@ def report_openvas_timestamp_scan_host(
             )
             return
 
+    def openvas_process_is_alive(
```
bjoernricks, line 1119: The name of the method should be
```suggestion
    def is_openvas_process_alive(
```

ospd_openvas/daemon.py
```diff
@@ -1116,6 +1116,38 @@ def report_openvas_timestamp_scan_host(
...
+        if parent and parent.status() == psutil.STATUS_ZOMBIE:
+            is_zombie = True
+
+        if (not parent_exists or is_zombie) and kbdb:
+            if kbdb and kbdb.scan_is_stopped(openvas_scan_id):
+                return True
+            kbdb.stop_scan(openvas_scan_id)
+            for scan_db in kbdb.get_scan_databases():
+                self.main_db.release_database(scan_db)
+            self.main_db.release_database(kbdb)
```
bjoernricks, line 1146: This doesn't fit to a method that checks something and returns a value. If a method has a side effect (that means it does change some data) it should be clearly visible from the method name and signature. In this case i would move this code to another method and call this method if is_openvas_process_alive(...) has returned true.

ospd_openvas/daemon.py
```diff
@@ -1263,9 +1295,19 @@ def exec_scan(self, scan_id: str):
 
         no_id_found = False
         while True:
+            if not self.openvas_process_is_alive(
+                kbdb, ovas_pid, openvas_scan_id
+            ):
+                logger.error(
+                    'Task %s was unexpectedly stopped or killed.', scan_id,
+                )
+                return 1
```
bjoernricks, line 1304: Just saw that the `exec_scan(self, scan_id: str):` signature is wrong. It must be `exec_scan(self, scan_id: str) -> int:` and needs a docstring about the meaning of the values but besides that returning an int from a function is not very pythonic. If you need to return some status code it is better to introduce an Enum for this purpose. If values > 0 are errors an exception should be raised instead.

ospd_openvas/daemon.py
```diff
@@ -1263,9 +1295,19 @@ def exec_scan(self, scan_id: str):
 
         no_id_found = False
         while True:
+            if not self.openvas_process_is_alive(
```
bjoernricks, line 1298:
```suggestion
            if not self.is_openvas_process_alive(
```

ospd_openvas/daemon.py
jjnicola: I agree, but this is out of the scope of this PR. I will prepare a new PR for this later

PR #236, opened 2020-04-16
Remove the VT cache and get the data directly from redis

jjnicola:


ospd_openvas/daemon.py
```diff
@@ -247,6 +253,52 @@ def format_vt_modification_time(self, value: str) -> str:
...
+
+                _elem_val = vt.get(_element)
+                _val = self.format_filter_value(_element, _elem_val)
+
+                if self.filter_operator[_oper](_val, _filter_val):
+                    continue
+                else:
+                    vt_oid_list.remove(vt_oid)
+
+        return vt_oid_list
```
bjoernricks, line 300: I am really sure this code can be simplified by using the [filter function](https://docs.python.org/3/library/functions.html#filter). This would avoid having to copy the list.

ospd_openvas/daemon.py
```diff
@@ -247,6 +253,52 @@ def format_vt_modification_time(self, value: str) -> str:
...
+            return None
+
+        if not self.nvti:
+            return None
+
+        vt_oid_list = [vtlist[1] for vtlist in self.nvti.get_oids()]
+        vt_oid_list_temp = copy.copy(vt_oid_list)
+        vthelper = VtHelper(self.nvti)
+
+        for _element, _oper, _filter_val in filters:
```
bjoernricks, line 282: Just a hint. There is no need to use underscore names here. The underscore names in loops should only indicate that this variable isn't used within the for loop (to make the linters happy too). E.g.

```python
for _id, value in some_list:
    do_something_with_the_value(value)
```

Additional info. In classes underscores are used to indicate private methods or variables of the class that could be accessed by other code technically but shouldn't.

```python
class Foo:
  def __init__(self):
    self._private_var = 'foo'

  def _private_method(self):
    self.do_something()
```

ospd_openvas/preferencehandler.py
```diff
@@ -126,18 +131,22 @@ def _get_vts_in_groups(self, filters: List[str],) -> List[str]:
...
-        vt_params_list = self.vts_cache[vtid].get("vt_params")
+        vt_params_list = vt.get("vt_params")
         if vt_params_list.get(vt_param_id):
             return vt_params_list[vt_param_id]["type"]
         return None
 
-    def _get_vt_param_name(self, vtid: str, vt_param_id: str,) -> Optional[str]:
+    def _get_vt_param_name(
+        self, vtid: str, vt_param_id: str, vt: Dict
+    ) -> Optional[str]:
```
bjoernricks, line 146:
```suggestion
    @staticmethod
    def _get_vt_param_name(
        vt: Dict, vtid: str, vt_param_id: str
    ) -> Optional[str]:
```

ospd_openvas/preferencehandler.py
```diff
@@ -126,18 +131,22 @@ def _get_vts_in_groups(self, filters: List[str],) -> List[str]:
 
         return vts_list
 
-    def _get_vt_param_type(self, vtid: str, vt_param_id: str) -> Optional[str]:
+    def _get_vt_param_type(
+        self, vtid: str, vt_param_id: str, vt: Dict
+    ) -> Optional[str]:
```
bjoernricks, line 136:
```suggestion
    @staticmethod
    def _get_vt_param_type(
        vt: Dict, vtid: str, vt_param_id: str
    ) -> Optional[str]:
```

ospd_openvas/preferencehandler.py
```diff
@@ -181,25 +190,30 @@ def _process_vts(
...
+                    vtid,
                 )
                 continue
 
             vts_list.append(vtid)
             for vt_param_id, vt_param_value in vt_params.items():
-                param_type = self._get_vt_param_type(vtid, vt_param_id)
-                param_name = self._get_vt_param_name(vtid, vt_param_id)
+                param_type = self._get_vt_param_type(vtid, vt_param_id, vt)
+                param_name = self._get_vt_param_name(vtid, vt_param_id, vt)
```
bjoernricks, line 211:
```suggestion
                param_type = self._get_vt_param_type(vt, vtid, vt_param_id)
                param_name = self._get_vt_param_name(vt, vtid, vt_param_id)
```

ospd_openvas/vthelper.py
```diff
@@ -0,0 +1,182 @@
...
+
+from ospd_openvas.nvticache import NVTICache
+
+
+class VtHelper:
+    def __init__(self, nvticache: NVTICache):
+        self.nvti = nvticache
+
+    def get_single_vt(self, vt_id: str, oids=None) -> Optional[Dict[str, any]]:
+        _custom = self.nvti.get_nvt_metadata(vt_id)
```
bjoernricks, line 33: As before. No need to user underscore variable names in this method.

ospd_openvas/daemon.py
jjnicola: We need to do a copy because the original list will change during the iteration. Doing the original list smaller (removing the vts that do not match), speeds up the algorithm. And using the copy for the iteration, allows to use more than one filter over the already filtered list:
e.g. modification_time > and < than date
"<get_vts filter='modification_time&gt;202004140211;modification_time&lt;202004150211' details='1'/>"

Given the case we want to allow the use of only one filter, I tested with filter, removing the list copy and writing the valid vts in a new list. This takes longer (38 seconds with filter() against 17 second with the current algorithm).

The memory peak during the copy is ~6 MB for some seconds
According to my tests, implementing the filter fun

bjoernricks: Thanks for the investigation. Seems your solution is better then the filter one.

PR #237, opened 2020-04-16
Fix test isolation

bjoernricks:
Ensure the mocks on Openvas.get_settings get removed after the test
runs.


tests/test_preferencehandler.py
```diff
@@ -527,16 +525,14 @@ def test_set_alive_pinghost(self, mock_kb):
         w.scan_collection.get_target_options = MagicMock(return_value=t_opt)
 
         ov_setting = {'some_setiting': 1}
```
cfi-gb, line 527:
```suggestion
        ov_setting = {'some_setting': 1}
```

Not directly related to this PR but could be fixed anyway.

PR #241, opened 2020-04-23
Check if cache is available

jjnicola:


bjoernricks:
I've restarted the failed test. Seems to be an issue with CircleCI itself.

Would be good to get tests for this too!

PR #247, opened 2020-04-30
Always send ALIVE_TEST when getting alive_test.

janowagner:
In case the alive_test is set for the target, then
always copy it over to the ALIVE_TEST scan preference.
Before, it was only send when also the setting
test_alive_hosts_only was true.

This helps openvas to decide on its own about using
ALIVE_TEST or not.


janowagner:
The actual challenge is solved meanwhile in a different way, so closing.

PR #252, opened 2020-05-06
Modify handling of dead hosts.

jjnicola:
Depends on PR greenbone/ospd#266

A finished host could be alive or dead, when the classic alive test method is uses.
Depending on the host progress it will be set as dead or alive(finished).

When Boreas is used, do not send log and host details message of dead hosts to the client.

In any case, the dead host are not taken in account for the scan progress calculation

### DIFF
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -809,12 +809,14 @@ class OSPDopenvas(OSPDaemon):
             launched, total = msg.split('/')
         except ValueError:
             return
+
         if float(total) == 0:
             return
         elif float(total) == -1:
-            host_prog = 100
+            host_prog = -1  # Host dead
         else:
             host_prog = (float(launched) / float(total)) * 100
+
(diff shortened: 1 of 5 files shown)


janowagner:
I tried the patch and it accelerates my Host Discovery scan substantially.

ospd_openvas/daemon.py
```diff
@@ -809,12 +809,14 @@ def update_progress(self, scan_id: str, current_host: str, msg: str):
...
         except ValueError:
             return
+
         if float(total) == 0:
             return
         elif float(total) == -1:
-            host_prog = 100
+            host_prog = -1  # Host dead
         else:
             host_prog = (float(launched) / float(total)) * 100
```
bjoernricks, line 818: Would be nice to use (or reuse) constants for the different progress values. Additionally calling `float` can fail e.g. `float(None)`. It would be good to catch the TypeError and do some safe error handling.

PR #257, opened 2020-05-14
Use flock for the feed lock file instead of just looking for existence.

jjnicola:


ospd_openvas/lock.py
```diff
@@ -29,35 +30,41 @@ class LockFile:
...
             parent_dir = self._lock_file_path.parent
             parent_dir.mkdir(parents=True, exist_ok=True)
 
-            self._lock_file_path.touch(exist_ok=False)
-            self._has_lock = True
-
-            logger.debug("Created lock file %s.", str(self._lock_file_path))
-
-        except FileExistsError as e:
+            self._fd = self._lock_file_path.open('w')
```
bjoernricks, line 49: I think we need to handle other exepction here to and always try to close the fd. also the fd should be set to none in case of errors.

ospd_openvas/lock.py
```diff
@@ -29,35 +30,41 @@ class LockFile:
...
                 e,
             )
+            return self
+
+        # Try to adquire the lock.
+        try:
+            fcntl.flock(self._fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+            self._has_lock = True
+            logger.debug("Created lock file %s.", str(self._lock_file_path))
+        except BlockingIOError as e:
```
bjoernricks, line 63: Same here. We should close the fd for all errors and set it to none.
something like
```python
except Exception as e:
    try:
        self._fd.close()
    except:
        pass
    raise e
```
not sure if this is completely correct

ospd_openvas/lock.py
```diff
@@ -71,10 +78,13 @@ def wait_for_lock(self):
     def release_lock(self) -> None:
         """ Release the lock by deleting the lock file
         """
-        if self.has_lock() and self.is_locked():
-            self._lock_file_path.unlink()
+        if self.has_lock() and self._fd:
+            fcntl.flock(self._fd, fcntl.LOCK_UN)
+            self._fd.close()
             self._has_lock = False
```
bjoernricks, line 84:
```suggestion
            self._fd = None
            self._has_lock = False
```

ospd_openvas/lock.py
```diff
@@ -29,56 +30,75 @@ class LockFile:
...
+        except Exception as e:  # pylint: disable=broad-except
             logger.error(
                 "Failed to create lock file %s. %s",
                 str(self._lock_file_path),
                 e,
             )
+            try:
+                self._fd.close()
+                self._fd = None
+            except (AttributeError, TypeError):
```
bjoernricks, line 59: Can't `self._fd.close()` raise other errors then AttributeError or TypeError ?

ospd_openvas/lock.py
```diff
@@ -29,56 +30,75 @@ class LockFile:
...
+                pass
+            return self
+
+        # Try to adquire the lock.
+        try:
+            fcntl.flock(self._fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+            self._has_lock = True
+            logger.debug("Created lock file %s.", str(self._lock_file_path))
+        except BlockingIOError as e:
```
bjoernricks, line 68: Is BlockingIOError the only error that can happen here?

PR #260, opened 2020-05-15
Support scan config default for Boreas

ArnoStiefvater:
Scan Config Default is an implicit alive_test method with value 0. If alive_test is 0 then only the defaults from ping_host.nasl are used. ICMP is set as default in ping_host.nasl. Other methods can be used by overwriting the defaults of ping_host.nasl. Therefore no explicit value for "Scan Config Default" was ever set because it was not needed.

Boreas (new method for alive detection) has no default alive_test set internally therefore an explicit alive_test is always needed.
Now ICMP is set for alive_test==0. This way "Scan Config Default" can be simulated.

### DIFF
--- a/ospd_openvas/preferencehandler.py
+++ b/ospd_openvas/preferencehandler.py
@@ -48,7 +48,7 @@ OID_SNMP_AUTH = "1.3.6.1.4.1.25623.1.0.105076"
 OID_PING_HOST = "1.3.6.1.4.1.25623.1.0.100315"
 
 BOREAS_ALIVE_TEST = "ALIVE_TEST"
-BOREAS = "test_alive_hosts_only"
+BOREAS_SETTING_NAME = "test_alive_hosts_only"
 
 
 class AliveTest(IntEnum):
@@ -394,14 +394,14 @@ class PreferenceHandler:
             )
 
     def prepare_boreas_alive_test(self):
-        """ Set alive_test for Boreas if
-        test_alive_hosts_only scanner config was set"""
+        """ Set alive_test for Boreas if boreas scanner config
+        (BOREAS_SETTING_NAME) was set"""
         settings = Openvas.get_settings()
         alive_test = -1
 
         if settings:
-            test_alive_hosts_only = settings.get('test_alive_hosts_only')
-            if not test_alive_hosts_only:
+            boreas = settings.get(BOREAS_SETTING_NAME)
+            if not boreas:
                 return
             alive_test_str = self.target_options.get('alive_test')
             if alive_test_str is not None:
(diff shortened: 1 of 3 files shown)


ospd_openvas/preferencehandler.py
```diff
@@ -379,37 +383,52 @@ def build_alive_test_opt_as_prefs(
...
+                    logger.debug(
+                        'Alive test preference for Boreas not set. '
+                        'Invalid alive test value %s.',
+                        alive_test_str,
+                    )
+            # ALIVE_TEST_SCAN_CONFIG_DEFAULT if no alive_test provided
+            else:
+                alive_test = AliveTest.ALIVE_TEST_SCAN_CONFIG_DEFAULT
+
+        if 1 <= alive_test <= 31:
```
bjoernricks, line 420: What does this check mean? Please add a comment or even better add constants for 1 and 31 too. Btw. I am not sure if the check works. Maybe it must be `if <= alive_test and alive_test <= 31`

ospd_openvas/preferencehandler.py
```diff
@@ -47,10 +47,14 @@
 OID_SNMP_AUTH = "1.3.6.1.4.1.25623.1.0.105076"
 OID_PING_HOST = "1.3.6.1.4.1.25623.1.0.100315"
 
+BOREAS_ALIVE_TEST = "ALIVE_TEST"
+BOREAS = "test_alive_hosts_only"
```
bjoernricks, line 51:
```suggestion
BOREAS_SETTING_NAME = "test_alive_hosts_only"
```

ospd_openvas/preferencehandler.py
ArnoStiefvater: `alive_test` is a bit mask. If an `alive_test` was set then the max value will be 11111 (31, all alive_test chosen) and min 1 (only first test chosen). So legal values if an `alive_test` was set are between 1 and 31.

bjoernricks: At the moment this line is not readable to understand its purpose. So at least it needs a comment. Better even e.g. if it is used to check the validity of the alive_test value then extract it to a function e.g.
```python
def is_valid_alive_test(self, alive_test):
    return alive_test >= ALIVE_TEST_MIN_VALUE and alive_test <= ALIVE_TEST_MAX_VALUE
```

ArnoStiefvater: Sound good. I will do that in a different PR as the comparison is already made in other functions too.

tests/test_preferencehandler.py
```diff
@@ -438,11 +438,11 @@ def test_set_boreas_alive_test_with_settings(self, mock_kb):
 
             p.kbdb.add_scan_preferences.assert_not_called()
 
-        # BOREAS set but invalid alive_test.
+        # Boreas conifg setting set but invalid alive_test.
```
bjoernricks, line 441:
```suggestion
        # Boreas config setting set but invalid alive_test.
```

PR #282, opened 2020-07-06
Improvements for fetching results from redis.

jjnicola:


ospd_openvas/db.py
```diff
@@ -178,6 +178,20 @@ def get_last_list_item(ctx: RedisCtx, name: str) -> str:
 
         return ctx.rpop(name)
 
+    @staticmethod
+    def pop_list_items(ctx: RedisCtx, name: str) -> str:
```
bjoernricks, line 182: Isn't it
```suggestion
    def pop_list_items(ctx: RedisCtx, name: str) -> List[str]:
```

ospd_openvas/daemon.py
```diff
@@ -1154,7 +1159,14 @@ def exec_scan(self, scan_id: str):
...
                 return
 
-            time.sleep(1)
+            # Wait a second before trying to get result from redis if there
+            # was no results before.
+            # Otherwise, wait 50 msec to give access other process to redis.
+            got_results = False
+            if not got_results:
+                time.sleep(1)
+            time.sleep(0.05)
```
bjoernricks, line 1168: Shouldn't be this in an else branch?
```suggestion
            else:
                time.sleep(0.05)
```

ospd_openvas/daemon.py
```diff
@@ -856,17 +856,21 @@ def get_severity_score(self, vt_aux: dict) -> Optional[float]:
...
-        res = db.get_result()
+        all_results = db.get_result()
         res_list = ResultList()
         total_dead = 0
-        while res:
+        total_results = len(all_results)
+
+        for res in all_results:
+            if not res:
+                continue
```
bjoernricks, line 873:
```suggestion
                continue

```

PR #285, opened 2020-07-10
Get results from main kb.

jjnicola:
OpenVAS stores now the results in the main task DB and all results are stored in the same internal/results kb in the main kb.
This allows to get all results stored in redis for all target's hosts at once. Also, avoid to loop looking for results in each single host kb.
Because it does not check the host kbs, ospd-openvas does not release them anymore and this is done by OpenVAS, unless in edge cases in which a scan is stopped or OpenVAS dies unexpectedly, ospd-openvas does the redis cleanup.


jjnicola:
Depends on PR greenbone/openvas#550 and PR greenbone/gvm-libs#376

PR #296, opened 2020-07-21
Log warning if wrong PLUGIN_SET format supplied

ArnoStiefvater:


ospd_openvas/daemon.py
```diff
@@ -377,6 +377,12 @@ def feed_is_outdated(self, current_feed: str) -> Optional[bool]:
             return None
 
         current_feed = safe_int(current_feed)
+        if current_feed is None:
+            logger.warning(
+                "Wrong PLUGIN_SET format in plugins feed file %s. Format has to"
```
cfi-gb, line 382: Could it be possible that this happens if the file isn't found at all?

ArnoStiefvater: There is already a debug message a few lines above for this case. `logger.debug('Plugins feed file %s not found.', feed_info_file)`.
Should I change the debug level to warning?

cfi-gb: If ospd-openvas can't work without an existing / accessible `plugin_feed_info.inc` then i would raise this debug to warning.

ArnoStiefvater: I raised the level to warning now.

PR #302, opened 2020-07-31
Fix optimize_test preference

ArnoStiefvater:


ospd_openvas/daemon.py
```diff
@@ -132,13 +132,20 @@
...
-        'type': 'integer',
+        'type': 'boolean',
         'name': 'optimize_test',
-        'default': 5,
+        'default': 1,
         'mandatory': 0,
         'visible_for_client': True,
         'description': (
-            'By default, openvas does not trust the remote ' + 'host banners.'
+            'By default, openvas does not trust the remote host banners. '
```
cfi-gb, line 141: It's the other way around. optimize_test is 1/yes by default which means remote host banners and so on are trusted which can be disabled by setting optimize_test to 0/no.

ospd_openvas/daemon.py
```diff
@@ -132,13 +132,21 @@
...
-        'type': 'integer',
+        'type': 'boolean',
         'name': 'optimize_test',
-        'default': 5,
+        'default': 1,
         'mandatory': 0,
         'visible_for_client': True,
         'description': (
-            'By default, openvas does not trust the remote ' + 'host banners.'
+            'By default, optimize_test is enabled which means openvas does '
```
cfi-gb, line 141: I would suggest a text like this:

>By default, optimize_test is enabled which means openvas does trust the remote host banners and is only launching plugins against the services they have been designed to check. For example it will check a web server claiming to be IIS only for IIS related flaws but will skip plugins testing for Apache flaws, and so on. This default behavior is used to optimize the scanning performance and to avoid false positives. If you are not sure that the banners of the remote host have been tampered with, you can disable this option.

ArnoStiefvater: Thanks, I will change this text in the manpage of openvas too.

cfi-gb:
I can confirm that the "optimize_test" value is now showing up as yes/no on NASL side again.

PR #311, opened 2020-08-07
Make all ospd-openvas config files more consistent:

cfi-gb:
- Don't use a DEBUG level in the example config file
- Add various default and more sane config parameters
  to the environment file
- Pass the arguments of the environment file to the systemd
  service

This has various reasons:

1. We shouldn't tell the user to use `/var/run/ospd/ospd.sock` and/or `/var/run/ospd.sock` because there might be multiple OSPD scanners in the future
2. Similar to 1. for `/var/run/ospd.pid`
3. All other GVM components (gvmd, greenbone-nvt-sync) are expecting the `feed-update.lock` in `<install-prefix>/var/run` with GVM-20.08 (see https://github.com/greenbone/gvmd/pull/1059 and https://github.com/greenbone/openvas/pull/507) but the current default is `/var/run/ospd`


cfi-gb:
The main problem i'm seeing is that we currently can't use something like cmake to configure / set the path during setup time like it is done when compiling other components.

PR #313, opened 2020-08-10
Update systemd unit description and drop Alias from it.

cfi-gb:
Same as in https://github.com/greenbone/gvmd/pull/1240#discussion_r467903517 for the Alias.

The description was taken from https://github.com/greenbone/ospd-openvas/blob/master/docs/ospd-openvas.8

### DIFF
--- a/config/ospd-openvas.service
+++ b/config/ospd-openvas.service
@@ -1,5 +1,5 @@
 [Unit]
-Description=OSPD OpenVAS Wrapper (ospd-openvas)
+Description=OSPd based OpenVAS scanner (ospd-openvas)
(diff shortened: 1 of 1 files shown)


config/ospd-openvas.service
```diff
@@ -1,5 +1,5 @@
 [Unit]
-Description=OSPD OpenVAS
+Description=OSPD OpenVAS Wrapper (ospd-openvas)
```
bjoernricks, line 2: Maybe
```suggestion
Description=OSPd based OpenVAS scanner (ospd-openvas)
```
Personally I am not a fan of the term wrapper.

PR #333, opened 2020-09-07
Describe mandatory --lock-file-dir / lock_file_dir configuration

cfi-gb:
**What**:

Should be self-explanatory

**Why**:

See e.g. https://github.com/Atomicorp/openvas/issues/25#issuecomment-687884201

**How**:

Not required

**Checklist**:

- [ ] Tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,16 @@ Please follow the general installation guide for ospd-based scanners:
 
   <https://github.com/greenbone/ospd/blob/master/doc/INSTALL-ospd-scanner.md>
 
+### Mandatory configuration
+
+The `ospd-openvas` startup parameter `--lock-file-dir` or the `lock_file_dir` config
+parameter of the `ospd.conf` config file needs to point to the same location / path of
+the `gvmd` daemon and the `openvas` command line tool (Default: `<install-prefix>/var/run`).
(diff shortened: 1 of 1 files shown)


bjoernricks (commented):
Is this really mandatory? Personally I never expected that the lock file needs to be the same in gvmd an ospd-openvas and would consider this a big.

cfi-gb:
@timopollmeier and/or @jjnicola might be able to tell more but my last information was that all components should / need to use the very same lock file.

ramsaut:
If the installation prefix is not empty, ospd-openvas still tries to create the lock file in /var/run/ospd/feed-update.lock instead of e.g. /opt/gvm/var/run/ospd/feed-update.lock and the configuration option is mandatory for the sync to work, because the /var/run/ospd/ directory does not exist and is not created.

jjnicola:
Currently, to not overlap feed related process, the path must be the same one used by greenbone-nvt-sync and gvmd. Of course there is place for improvements. While some default configuration changes can be done in the near future, a more invasive change like reducing the dependency on the lock file of other modules will be possible schedule for the next mayor release.

jjnicola (approved):

PR #336, opened 2020-09-15
Improve logging for unsatisfied vts dependencies

jjnicola:
**What**:
Log the file name of the missing dependency.

**Why**:
Because it is currently log only the OID of the problematic VT, but not the name of the missing dependency.

**How**:
Modify a nasl script, adding a typo to an existing dependency. Run openvas -u to update the Redis cache. Run <get_vts/> command. You will see in the logs, the name of the missing deps for the modified nasl plugin.

**Checklist**:

- [x] Tests

### DIFF
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -359,6 +359,8 @@ OSPD_PARAMS = {
     },
 }
 
+VT_BASE_OID = "1.3.6.1.4.1.25623."
+
 
 def safe_int(value: str) -> Optional[int]:
     """Convert a string into an integer and return None in case of errors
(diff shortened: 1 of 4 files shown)


ospd_openvas/daemon.py
```diff
@@ -717,9 +717,9 @@ def get_dependencies_vt_as_xml_str(
         vt_deps_xml = Element('dependencies')
         for dep in vt_dependencies:
             _vt_dep = Element('dependency')
-            try:
+            if "1.3.6.1.4.1.25623." in dep:
```
bjoernricks, line 720: Could you please add a global constant for this OID value. From the outside this value is very difficult to understand.

bjoernricks (changes requested):

bjoernricks (approved):

PR #343, opened 2020-10-09
Handle HOSTS_COUNT message.

jjnicola:
**What**:
OpenVAS sent a message with the updated count and the message is processed.

**Why**:
Sometimes the amount of hosts to be scanned differs from the original
amount from the target string.

**How**:
Run a scan against a target with a hostname resolving to more than one IP (e.g. ipv4 and ipv6)

**Checklist**:

- [ ] Tests


jjnicola:
Depends on greenbone/ospd#332

bjoernricks (approved):

PR #346, opened 2020-10-20
Support CVSSv3 and use new tags for severity vector, origin, date.

jjnicola:
**What**:
If new tags are present, check also the type and set it.
Otherwise, it asumes cvss v2, don't set an origin and use the creation time for
the severity date element.

**Why**:
Because CVSSv3 support is required.

**How**:
Add to some script the following tags (in the example the script with OID:1.3.6.1.4.1.25623.1.0.100082):

```
  script_tag(name:"severity_score", value:"CVSS:3.0/AV:L/AC:H/PR:H/UI:R/S:U/C:N/I:L/A:L");
  script_tag(name:"severity_origin", value:"greenbone");
  script_tag(name:"severity_date", value:"1603191789");
```
To check the output 

```
gvm-cli --protocol OSP --timeout 120 socket --socketpath=/var/run/ospd/openvas.sock --xml "<get_vts vt_id='1.3.6.1.4.1.25623.1.0.100082'/>" |xmlstarlet fo
```

**Checklist**:

- [ ] Tests


mattmundell:
Works for me.

bjoernricks (approved):

PR #364, opened 2020-12-17
Check if Notus is enabled.

jjnicola:
**What**:
Uses the new scanner only option "table_driven_lsc" to check if
notus is enabled. If it is enabled, it loads the Notus metadata
into the redis cache.

**Why**:
To upload the Notus metadata if required

**How**:
Enable/disable Notus in the openvas configuration file and check the logs during plugins load up.

**Checklist**:

- [x] Tests


jjnicola:
Depends on greenbone/openvas#632

ArnoStiefvater (approved):

PR #374, opened 2021-01-25
Extend script_xref logging on unexpected data / input.

cfi-gb:
**What**:

Previously to this change we had the following log entry:

```
OSPD[9960] 2021-01-23 10:39:35,939: ERROR: (ospd_openvas.daemon) Not possible to parse xref Protocols and Services Are Running for VT 1.3.6.1.4.1.25623.1.0.150070
```

which had two issues:

1. There was no clear separation between the text of the log message and the text of the malformed xref.
2. The exception info was missing.

After this PR we now have:

```
OSPD[15404] 2021-01-25 09:02:42,784: ERROR: (ospd_openvas.daemon) Not possible to parse xref "Protocols and Services Are Running" for VT 1.3.6.1.4.1.25623.1.0.150070: not enough values to unpack (expected 2, got 1)
```

**Why**:

Better / more consistent logging.

**How**:

Startup `ospd-openvas` against the current feed with a VT that has the following `script_xref`:

```
script_xref(name:"Policy", value:"CIS Controls Version 7: 9.2 Ensure Only Approved Ports, Protocols and Services Are Running");
```

**Checklist**:

- [ ] Tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ### Changed
 - Improve logging for unsatisfied vts dependencies. [#336](https://github.com/greenbone/ospd-openvas/pull/336)
 - Do not use busy wait when waiting for the openvas scan process to finish. [#360](https://github.com/greenbone/ospd-openvas/pull/360)
+- Improve logging for unexpected data in script_xref tags. [#374](https://github.com/greenbone/ospd-openvas/pull/374)
 
 ### Fixed
 - Fix nvticache name when for stable version from sources. [#317](https://github.com/greenbone/ospd-openvas/pull/317)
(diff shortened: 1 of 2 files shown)


jjnicola (approved):

romaaw:
@cfi-gb, can you please clarify if such a log record is an error?
```
OSPD[15404] 2021-01-25 09:02:42,784: ERROR: (ospd_openvas.daemon) Not possible to parse xref "Protocols and Services Are Running" for VT 1.3.6.1.4.1.25623.1.0.150070: not enough values to unpack (expected 2, got 1)
```

What should I do to get rid of it?
Thanks.

jjnicola:
Hi @romaaw,
The problem is the usage of ',' (comma) in the xref's value string, which is currently not supported, and will be fixed soon in the script. Once the script is fixed, after a feed update, the message will disappear.
Regards,
Juan

cfi-gb:
Note that the topic above is / was completely unrelated and this PR just had extended the logging of this.

The related PR to extend `openvas-nasl-lint` to check for such currently unsupported chars is currently open at https://github.com/greenbone/openvas-scanner/pull/650 which will prevent that VTs using these are arriving in the feed.

PR #390, opened 2021-03-01
Retry host progress update if it the progress is still below 100 percent.

jjnicola:
**What**:
Check the scan progress of host up to three time if it
was considered finishes or dead and the progress is still under 100%
The retries could avoid a race condition in which a finished host still
has the progress udner 100%, which leads to interrupted scans.

**Why**:
The retries could avoid a race condition in which a finished host still
has the progress udner 100%, which leads to interrupted scans.

**How**:
This is quite difficult to test since some interrupted scans can be the result of a race condition. 

**Checklist**:

- [ ] Tests
- [X] [CHANGELOG](https://github.com/greenbone/gsa/blob/master/CHANGELOG.md) Entry

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,7 +33,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix alive test preferences when a non default method is selected. [#334](https://github.com/greenbone/ospd-openvas/pull/334)
 - Check for empty vts preferences list. [#340](https://github.com/greenbone/ospd-openvas/pull/340)
 - Fix progress calculation when the host count differs from the target string count. [#343](https://github.com/greenbone/ospd-openvas/pull/343)
-- Retry host progress update if it the progress is still below 100 percent. [#390](https://github.com/greenbone/ospd-openvas/pull/390)
+- Retry host progress update if the progress is still below 100 percent. [#390](https://github.com/greenbone/ospd-openvas/pull/390)
 
 [20.8.1]: https://github.com/greenbone/ospd-openvas/compare/v20.8.0...ospd-openvas-20.08
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix alive test preferences when a non default method is selected. [#334](https://github.com/greenbone/ospd-openvas/pull/334)
 - Check for empty vts preferences list. [#340](https://github.com/greenbone/ospd-openvas/pull/340)
 - Fix progress calculation when the host count differs from the target string count. [#343](https://github.com/greenbone/ospd-openvas/pull/343)
+- Retry host progress update if it the progress is still below 100 percent. [#390](https://github.com/greenbone/ospd-openvas/pull/390)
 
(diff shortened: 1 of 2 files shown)


jjnicola:
Depends on greenbone/ospd#363

ospd_openvas/daemon.py
```diff
@@ -1401,20 +1401,40 @@ def exec_scan(self, scan_id: str):
...
+                            host_progress > -1 and host_progress < 100
+                        ) and retry > 0:
+                            self.report_openvas_scan_status(
+                                scan_db, scan_id, current_host
+                            )
+                            time.sleep(1)
+                            retry -= 1
+                            host_progress = self.get_scan_host_progress(
+                                scan_id, current_host
+                            )
```
paulnaumann, line 1422: This will only partially solve this problem. 
1. Why was 3 (seconds) chosen? This will probably only reduce the number of scans marked as INTERRUPTED, won't it?
2. If retry reaches 0, the while loop will end, but there is no separate logic to handle this case. An error should be raised or at least the error should be logged, if the following code works fine for 'invalid'  host progress data

jjnicola: 1. Yes, the idea is to reduce the amount of "INTERRUPTED", based in the suspicion of a race condition in this part of the code. The host finished but the progress still not updated. It retries 3 times one second, which is different to wait 3 seconds. I think 1 second for each retries is enough time to let redis to be updated and get the last plugins launch rate 100/100. If it is a race condition, on the first retry it should be leave the while loop.
2. You are right. Solved with https://github.com/greenbone/ospd-openvas/pull/390/commits/11e833808a29c9f93a7111744a85408ae1b56960

Thanks for the review!

CHANGELOG.md
```diff
@@ -33,6 +33,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - Fix alive test preferences when a non default method is selected. [#334](https://github.com/greenbone/ospd-openvas/pull/334)
 - Check for empty vts preferences list. [#340](https://github.com/greenbone/ospd-openvas/pull/340)
 - Fix progress calculation when the host count differs from the target string count. [#343](https://github.com/greenbone/ospd-openvas/pull/343)
+- Retry host progress update if it the progress is still below 100 percent. [#390](https://github.com/greenbone/ospd-openvas/pull/390)
```
ArnoStiefvater, line 36:
```suggestion
- Retry host progress update if the progress is still below 100 percent. [#390](https://github.com/greenbone/ospd-openvas/pull/390)
```

ospd_openvas/daemon.py
```diff
@@ -1401,20 +1401,66 @@ def exec_scan(self, scan_id: str):
...
+                            time.sleep(1)
+                            self.report_openvas_scan_status(
+                                scan_db, scan_id, current_host
+                            )
+                            retry -= 1
+                            host_progress = self.get_scan_host_progress(
+                                scan_id, current_host
+                            )
 
+                        if retry == 0:
```
ArnoStiefvater, line 1432: Even if after the third retry we successfully get a valid host_progress the `retry` will be set to 0 nonetheless and an error is logged. Maybe a `break` can be used in the while loop.

ArnoStiefvater (changes requested):

ArnoStiefvater (approved):

PR #412, opened 2021-04-26
Fix timezone

Kraemii:
**What**:
The logger now uses UTC for timestamps.

**Why**:
Other tools used UTC and ospd-openvas used local time.

**How**:
Adding one line in which the timezone of the logger is set to UTC. Simply running ospd-openvas showed that this change works.

**Checklist**:

- [ ] Tests N/A


Kraemii:
Closing in favor of (https://github.com/greenbone/ospd/pull/394)

PR #449, opened 2021-06-28
Move MQTT feature into new branch

ArnoStiefvater:
**What**:

Reintroduce https://github.com/greenbone/ospd-openvas/pull/422 after rebase into new branch `middleware`.

**Why**:

Separation of feature branch and master.

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -1006,16 +1006,6 @@ class OSPDopenvas(OSPDaemon):
         """Get all results from redis kb and add them into the scan table"""
         all_results = db.get_result()
 
-        return self.report_openvas_results_redis_format_to_dict(
-            scan_id, all_results
-        )
-
-    def report_openvas_results_redis_format_to_dict(
-        self, scan_id: str, all_results: list
-    ) -> bool:
-        """Transforms all Results from redis format into a dictionary format
-        and add them into the scan table
-        """
         results = []
         for res in all_results:
             if not res:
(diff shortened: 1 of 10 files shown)


ospd_openvas/daemon.py
```diff
@@ -980,36 +1002,71 @@ def report_openvas_scan_status(self, kbdb: BaseDB, scan_id: str):
...
+            result["hostname"] = msg[2]
+            result["port"] = msg[3]
+            result["OID"] = msg[4]
+            result["value"] = msg[5]
+            if len(msg) > 6:
+                result["uri"] = msg[6]
+
+            results.append(result)
+
+        return self.report_openvas_results(results, scan_id)
```
Kraemii, line 1036:
```suggestion
    def report_openvas_results_redis(self, db: BaseDB, scan_id: str) -> bool:
        """Get all results from redis kb and add them into the scan table"""
        all_results = db.get_result()

        results = []
        for res in all_results:
            if not res:
                continue
            result = {}
            msg = res.split('|||')
            result["type"] = msg[0]
            result["host_ip"] = msg[1]
            result["hostname"] = msg[2]
            result["port"] = msg[3]
            result["OID"] = msg[4]
            result["value"] = msg[5]
            if len(msg) > 6:
                result["uri"] = msg[6]

            results.append(result)

        return self.report_openvas_results(results, scan_id)
```
Combine `report_openvas_results_redis` and `report_openvas_results_redis_format_to_dict` as there were only separated for testing reasons when results sent with mqtt also had the redis format.

Kraemii (approved):

PR #493, opened 2021-08-19
Set required version of psutil in setup.py

simonflood:
**What**:

Set requirement for Python psutil module to be >=5.7.2,<6.0.0 in setup.py.

**Why**:

Following the docs to build GVM 21.04 on Ubuntu 20.04, installing ospd-openvas reports this requirement as an error.

**How**:

On Ubuntu 20.04 I manually made this change via sed, then re-ran the relevant "python3 -m pip install" command.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,7 +50,7 @@ setup(
     install_requires=[
         'ospd>=21.10.0.dev1',
         'redis>=3.0.1',
-        'psutil',
+        'psutil>=5.7.2,<6.0.0',
         'packaging',
     ],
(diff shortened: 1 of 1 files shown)


bjoernricks:
Hi thanks for contributing. This issue has already been addressed in https://github.com/greenbone/ospd-openvas/pull/490 and will be fixed with the next release. Actually we require a lower version of psutil and have adjusted our the setup file to use a version >= 5.5.1 which is the version shipped in debian buster.

PR #510, opened 2021-10-20
Integrate notus

Kraemii:
**What**:
Integrate Notus into ospd-openvas to get results.
SC-430 SC-431

**Why**:
To support the notus scanner

**How**:
Tested it with gvm-cli and the following script:
```xml
<start_scan scan_id='97079ee9-8917-49da-aa4f-4ef95f757ac1'>
    <vt_selection>
        <vt_single id='1.3.6.1.4.1.25623.1.0.50282'>
        </vt_single>
    </vt_selection>
    <targets>
        <target>
            <hosts>192.168.9.39</hosts>
            <ports>80,8080,443,22</ports>
            <credentials>
                <credential service="ssh" type="up" port="22">
                    <username>user</username>
                    <password>password</password>
                </credential>
            </credentials>
        </target>
    </targets>
    <scanner_params>
    </scanner_params>
</start_scan>
```
The vt will gather the package list and will start the table driven lsc (when activated). The host must be a supported EulerOS or Suse system.

**Checklist**:

- [x] Tests


ospd_openvas/messaging/mqtt.py
```diff
@@ -0,0 +1,159 @@
...
+
+    @staticmethod
+    def on_connect(_client, _userdata, _flags, rc, _properties):
+        if rc == 0:
+            logger.info("Connected to broker successfully")
+        else:
+            logger.error('Failed to connect to broker. Reason code %s', rc)
+
+    @staticmethod
+    def on_disconnect(client, _userdata, rc=0):
```
y0urself, line 154:
```suggestion
    def on_disconnect(client: MQTTClient, _userdata, rc: int = 0) -> None:
```
I am not sure what userdata is for, so you may fill it on your own ...?

ospd_openvas/messaging/mqtt.py
```diff
@@ -0,0 +1,159 @@
...
+        client: MQTTClient,
+    ):
+        self._client = client
+        self._client.on_disconnect = self.on_disconnect
+        self._client.on_connect = self.on_connect
+
+        self._client.connect()
+
+    @staticmethod
+    def on_connect(_client, _userdata, _flags, rc, _properties):
```
y0urself, line 147:
```suggestion
    def on_connect(_client: MQTTClient, _userdata, _flags, rc: int, _properties) -> None:
```
The other fields are not clear for me.

Edit: Urgh. Sorry, paho is still python2 compatible and its python code looks really aweful because of that. :D

ospd_openvas/notus.py
```diff
@@ -0,0 +1,54 @@
...
+
+    def result_handler(self, res_msg: ResultMessage) -> None:
+
+        result = res_msg.serialize()
+        scan_id = result.pop("scan_id")
+        timer = None
+
+        if not scan_id in self._results:
+            self._results[scan_id] = []
+            timer = Timer(0.25, self.report_results, [scan_id])
```
y0urself, line 50: how about using a parameter for the timer value in case we want to check out, what value is best? (or a config file value)?

ospd_openvas/daemon.py
```diff
@@ -980,36 +1030,62 @@ def report_openvas_scan_status(self, kbdb: BaseDB, scan_id: str):
 
         self.sort_host_finished(scan_id, finished_hosts)
 
+    def report_notus_results(self, results: list, scan_id: str) -> bool:
+        return self.report_results(results, scan_id)
```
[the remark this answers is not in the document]
Kraemii: clear naming, if we use `report_openvas_results` for openvas we should use `report_notus_results` for notus

ospd_openvas/daemon.py
```diff
@@ -1360,6 +1439,8 @@ def exec_scan(self, scan_id: str):
                 time.sleep(0.05)
             got_results = False
 
+        # Sleep a second to be sure to get all notus results
+        time.sleep(1)
```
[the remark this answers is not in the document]
Kraemii: The problem is, that we wait 0.25 seconds after receiving a notus result, before publishing all results we got in this time. I tested it without this sleep and sometimes I did not get all notus results. I don't think this will be a problem at all, as we simply wait another second at the end of a scan, which normally take some minutes to several hours.

ospd_openvas/messaging/mqtt.py
```diff
@@ -0,0 +1,159 @@
...
+        keepalive=60,
+        bind_address="",
+        bind_port=0,
+        clean_start=mqtt.MQTT_CLEAN_START_FIRST_ONLY,
+        properties=None,
+    ):
+        if not host:
+            host = self._mqtt_broker_address
+        if not port:
+            port = self._mqtt_broker_port
```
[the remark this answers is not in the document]
Kraemii: ??

ospd_openvas/messaging/mqtt.py
Kraemii: userdata can be anything or nothing, you can give the function any additional information you like within the client itself. You can set it with something like
```py
self._client.user_data_set()
```
or
```py
self._client._userdata = "foo"
```
therefore I don't think it makes any sense to set any type for it. Also I thing it is not necessary to specify any type to the `on_`-functions as they are not intended to be called by a user.

ospd_openvas/notus.py
Kraemii: IMHO I don't think the timer value matters that much and I don't like configuring this time, as there can also happen bugs, if someone set this time too high. So introducing a config for something that is not user relevant, but can cause bugs when set manually sounds dangerous. If we wan to find out the best value for this, the developer can test it on their own by changing this value(?)

ospd_openvas/messaging/mqtt.py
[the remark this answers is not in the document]
Kraemii: That would be something the developer is responsible to. In this case he should do a try catch block, as the only thing I can add would be throwing an Exception if something uninitialized and that already happens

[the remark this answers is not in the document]
y0urself: I meant, that in `__init__` you don't take care of `mqtt_broker_address`and `mqtt_broker_port` to be initialized? So these values may be `None` causing an error?

ospd_openvas/messaging/mqtt.py
y0urself: yeah. See my edit ... I looked into paho ...

ospd_openvas/messaging/mqtt.py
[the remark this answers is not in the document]
Kraemii: Yes that might be, but what should I do about this? Check it and raise an exception? Or print a log message? The exception will come either way and has to be handled by the caller.

ospd/parser.py
```diff
@@ -170,24 +175,50 @@ def __init__(self, description: str) -> None:
...
+            help=(
+                'Broker address to connect to for MQTT communication.'
+                ' Neccessary to get results from Notus-Scanner.Default'
+                ' %(default)s'
+            ),
+        )
+        parser.add_argument(
+            '--mqtt-broker-port',
+            default=DEFAULT_MQTT_BROKER_PORT,
+            type=int,
```
y0urself, line 216:
```suggestion
            type=self.network_port,
```

ospd_openvas/messaging/mqtt.py
```diff
@@ -0,0 +1,160 @@
...
+
+        return super().connect(
+            host,
+            port=port,
```
y0urself, line 69:
```suggestion
            port=int(port), // because of config, feeling cute, might delete later.
```

Kraemii: The parser already parses network ports to int:
```py
    def network_port(self, string: str) -> int:
        """Check if provided string is a valid network port."""

        value = int(string)
        if not 0 < value <= 65535:
            raise argparse.ArgumentTypeError(
                'port must be in ]0,65535] interval'
            )
        return value
```

y0urself: yeah but also if gathered from config file?

Kraemii: At least for me :trollface: 
I updated it, so you can test it

ospd_openvas/messaging/mqtt.py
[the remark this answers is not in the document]
y0urself: After understanding the default value thingy, I think we can assume that these values are always set to the default ...

y0urself (approved):

PR #518, opened 2021-10-28
Add: notus advisories in get_vts

nichtsfrei:
When calling get_vts ospd-openvas will try to load the notus advisories;
look them up and yielding the response; when it is not in the notus
advisories it will return the information based on redis from openvas as
previously.

With this integration a client e.g. gvmd should get all needed
information for notus results.

**What**:

**Why**:

**How**:

**Checklist**:

- [ ] Tests
- [ ] PR merge commit message adjusted

### DIFF
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -432,9 +432,9 @@ class OSPDopenvas(OSPDaemon):
     ):
         """Initializes the ospd-openvas daemon's internal data."""
         self.main_db = MainDB()
-        notus_dir = kwargs.get( 'notus_feed_dir')
+        notus_dir = kwargs.get('notus_feed_dir')
         notus = Notus(notus_dir) if notus_dir else None
-        
+
         self.nvti = NVTICache(
             self.main_db,
             notus,
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -23,8 +23,6 @@
 
 import logging
 
-from ospd.parser import CliParser
-from ospd_openvas.notus import Notus
 import time
 import copy
 
@@ -37,6 +35,7 @@ from lxml.etree import tostring, SubElement, Element
 
 import psutil
 
+from ospd.parser import CliParser
 from ospd.ospd import OSPDaemon
 from ospd.scan import ScanProgress, ScanStatus
 from ospd.server import BaseServer
@@ -47,6 +46,7 @@ from ospd.resultlist import ResultList
 from ospd_openvas import __version__
 from ospd_openvas.errors import OspdOpenvasError
 
+from ospd_openvas.notus import Notus
 from ospd_openvas.dryrun import DryRun
 from ospd_openvas.nvticache import NVTICache
 from ospd_openvas.db import MainDB, BaseDB
@@ -432,13 +432,12 @@ class OSPDopenvas(OSPDaemon):
     ):
         """Initializes the ospd-openvas daemon's internal data."""
         self.main_db = MainDB()
+        notus_dir = kwargs.get( 'notus_feed_dir')
+        notus = Notus(notus_dir) if notus_dir else None
+        
         self.nvti = NVTICache(
             self.main_db,
-            Notus(
-                kwargs.get(
-                    'notus_feed_dir', '/var/lib/openvas/notus/advisories'
-                )
-            ),
+            notus,
         )
 
         super().__init__(
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -433,7 +433,7 @@ class OSPDopenvas(OSPDaemon):
         """Initializes the ospd-openvas daemon's internal data."""
         self.main_db = MainDB()
         notus_dir = kwargs.get('notus_feed_dir')
-        notus = Notus(notus_dir) if notus_dir else None
+        notus = Notus(notus_dir, self.main_db.ctx) if notus_dir else None
 
         self.nvti = NVTICache(
             self.main_db,
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -22,6 +22,9 @@
 """ Setup for the OSP OpenVAS Server. """
 
 import logging
+
+from ospd.parser import CliParser
+from ospd_openvas.notus import Notus
 import time
 import copy
 
@@ -429,7 +432,7 @@ class OSPDopenvas(OSPDaemon):
     ):
         """Initializes the ospd-openvas daemon's internal data."""
         self.main_db = MainDB()
-        self.nvti = NVTICache(self.main_db)
+        self.nvti = NVTICache(self.main_db, Notus(kwargs['notus_feed_dir']))
 
         super().__init__(
             customvtfilter=OpenVasVtsFilter(self.nvti),
@@ -1365,9 +1368,20 @@ class OSPDopenvas(OSPDaemon):
         self.main_db.release_database(kbdb)
 
 
+class NotusParser(CliParser):
+    def __init__(self):
+        super().__init__('OSPD - openvas')
+        self.parser.add_argument(
+            '--notus-feed-dir',
+            default="/var/lib/openvas/notus/advisories",
+            help='Directory where notus feed is placed. Default: %(default)s',
+        )
+
+
 def main():
     """OSP openvas main function."""
-    daemon_main('OSPD - openvas', OSPDopenvas)
+
+    daemon_main('OSPD - openvas', OSPDopenvas, NotusParser())
 
 
 if __name__ == '__main__':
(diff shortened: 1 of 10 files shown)


ospd_openvas/vthelper.py
```diff
@@ -20,17 +20,25 @@
 """ Provide functions to handle VT Info. """
 
 from hashlib import sha256
+import logging
 from typing import Optional, Dict, List, Tuple, Iterator
 
+from typing import Any
```
bjoernricks, line 26:
```suggestion
from typing import Any, Optional, Dict, List, Tuple, Iterator

```

ospd_openvas/notus.py
```diff
@@ -0,0 +1,107 @@
...
+        for key, value in self.cache.items():
+            yield (Path(value).name, key)
+
+    def find_advisory(self, path: Path, oid: str):
+        advisories = json.loads(path.read_bytes())
+        for advisory in advisories.get("advisories", []):
+            if advisory.get("oid") == oid:
+                return advisory
+
+    def get_nvt_metadata(self, oid: str) -> Optional[Dict[str, str]]:
```
bjoernricks, line 60: I really don't like putting this kind of structured data into a dict. This should be stored in a dataclass instead. But I guess this is also already the case with the openvas metadata.

nichtsfrei: yes, that's already the case for openvas meta data so I would not like to change that; besides that aren't data classes pretty much unusable by a serializer (like xml in this case)?

Kraemii:
For me it took about 20 mins to do the integrity check. Also the OSP get_vts command takes about 20 mins. In comparison the integrity check as well as the get_vts command with the whole vts from the community feed takes about 1 min.

Kraemii:
Many information are missing for the vts from the notus feed when calling get_vts. Comparison:
```xml
<vt id="1.3.6.1.4.1.25623.1.1.2.2021.2561">
      <name>Huawei EulerOS: Security Advisory for polkit (EulerOS-SA-2021-2561)</name>
      <creation_time>1632812890</creation_time>
      <modification_time>1632812890</modification_time>
      <summary>The remote host is missing an update for the Huawei EulerOS 'polkit' package(s) announced via the EulerOS-SA-2021-2561 advisory.</summary>
      <affected>'polkit' package(s) on Huawei EulerOS V2.0SP9(x86_64).</affected>
      <insight>polkit is a toolkit for defining and handling authorizations. It isused for allowing unprivileged processes to speak to privilegedprocesses.(CVE-2021-3560)</insight>
</vt>
```
```xml
<vt id="1.3.6.1.4.1.25623.1.1.2.2021.2561">
      <name>Huawei EulerOS: Security Advisory for polkit (EulerOS-SA-2021-2561)</name>
      <refs>
        <ref id="CVE-2021-3560" type="cve"/>
        <ref id="EulerOS-SA-2021-2561" type="advisory-id"/>
        <ref id="https://developer.huaweicloud.com/ict/en/site-euleros/euleros/security-advisories/EulerOS-SA-2021-2561" type="url"/>
      </refs>
      <dependencies>
        <dependency vt_id="1.3.6.1.4.1.25623.1.0.143355"/>
      </dependencies>
      <creation_time>1632812890</creation_time>
      <modification_time>1632812890</modification_time>
      <summary>The remote host is missing an update for the Huawei EulerOS 'polkit' package(s) announced via the EulerOS-SA-2021-2561 advisory.</summary>
      <affected>'polkit' package(s) on Huawei EulerOS V2.0SP9(x86_64).</affected>
      <insight>polkit is a toolkit for defining and handling authorizations. It isused for allowing unprivileged processes to speak to privilegedprocesses.(CVE-2021-3560)</insight>
      <solution type="VendorFix">Please install the updated package(s).</solution>
      <detection qod_type="package">Checks if a vulnerable package version is present on the target host.</detection>
      <severities>
        <severity type="cvss_base_v3">
          <value>CVSS:3.1/AV:N/AC:L/PR:N/UI:N/S:U/C:L/I:N/A:N</value>
          <origin>Greenbone</origin>
          <date>1632812876</date>
        </severity>
      </severities>
      <custom>
        <filename>2021/huawei/gb_euleros_sa_2021_2561.nasl</filename>
        <mandatory_keys>ssh/login/euleros, ssh/login/rpms, ssh/login/release=EULEROS\-2\.0SP9\-X86_64</mandatory_keys>
        <cvss_base_vector>AV:N/AC:L/Au:N/C:P/I:N/A:N</cvss_base_vector>
        <category>3</category>
        <family>Huawei EulerOS Local Security Checks</family>
      </custom>
</vt>
```
Some things are clear, as we do not have any dependencies within notus, but the severity and CVEs are missing completely. Also the solution should always be
```xml
<solution type="VendorFix">Please install the updated package(s).</solution>
```

Kraemii (dismissed):
For me it took about 20 mins to do the integrity check. Also the OSP get_vts command takes about 20 mins. In comparison the integrity check as well as the get_vts command with the whole vts from the community feed takes about 1 min.

ospd_openvas/notus.py
```diff
@@ -0,0 +1,107 @@
...
+        if not self.cache:
+            self.reload_cache()
+        for key, value in self.cache.items():
+            yield (Path(value).name, key)
+
+    def find_advisory(self, path: Path, oid: str):
+        advisories = json.loads(path.read_bytes())
+        for advisory in advisories.get("advisories", []):
+            if advisory.get("oid") == oid:
+                return advisory
```
Kraemii, line 58: When getting all vts or doing the integrity check the program always loads the whole file (~11mb for euleros) for each vt. This causes the the integrity check and get_vts OSP command to take about 20 min. Alternatively save all nvt metadata within the nvt cache so reading the file has to be only done once at startup and each time the feed gets updated.

nichtsfrei:
For the missing solution as well as detection did miss since it isn't in the file and I actually asked about that in the channel; going to add those. I am a bit puzzled why in your case the severities, dependencies as well as custom are missing since they are there when I run the tests with eulaos.

Also going to add caching.

Kraemii:
The reason why solution and detection is missing is because it is the same for all notus advisories:
```xml
<solution type="VendorFix">Please install the updated package(s).</solution>
<detection qod_type="package">Checks if a vulnerable package version is present on the target host.</detection>
```

ospd_openvas/daemon.py
```diff
@@ -1365,9 +1374,20 @@ def exec_scan(self, scan_id: str):
...
 
 
+class NotusParser(CliParser):
+    def __init__(self):
+        super().__init__('OSPD - openvas')
+        self.parser.add_argument(
+            '--notus-feed-dir',
+            default="/var/lib/openvas/notus/advisories",
+            help='Directory where notus feed is placed. Default: %(default)s',
+        )
```
Kraemii, line 1384: I think it would be better to put it into an additional module or integrate it into the OSPD parser

ospd_openvas/notus.py
```diff
@@ -0,0 +1,122 @@
...
+        result["modification_time"] = str(advisory.get("last_modification", 0))
+        result["summary"] = advisory.get("summary")
+        result["impact"] = advisory.get("impact")
+        result["affected"] = advisory.get("affected")
+        result["insight"] = advisory.get("insight")
+        result['solution'] = "Please install the updated package(s)."
+        result['solution_type'] = "VendorFix"
+        result[
+            'vuldetect'
+        ] = 'Checks if a vulnerable package version is present on the target host.'
```
Kraemii, line 86:
```suggestion
        result['vuldetect'] = (
            'Checks if a vulnerable package version is present on the target'
            ' host.'
        )
```

ospd_openvas/nvticache.py
```diff
@@ -89,7 +92,13 @@ def get_oids(self) -> Iterator[Tuple[str, str]]:
...
             as first element and the oid as second one.
         """
-        return OpenvasDB.get_filenames_and_oids(self.ctx)
+        if self.notus:
+            for f, oid in self.notus.get_filenames_and_oids():
+                yield (f, oid)
+        if self.ctx:
+            for f, oid in OpenvasDB.get_filenames_and_oids(self.ctx):
+                if not self.notus or not self.notus.exists(oid):
+                    yield (f, oid)
```
Kraemii, line 101: What does the A i. in the docstring mean? I think the docstring is wrong as we return a Tuple, not a List

nichtsfrei: It is actually returning a generator / iterator of a tuple so list seems kind of matching to me. But going to change it

ospd_openvas/daemon.py
nichtsfrei: Can you elaborate your thought? It is very small so an extra module seems a bit hefty and putting it deep into OSPD into an already long seems a bit confusing to me.

Kraemii: It feels wrong to me to have the parser in the demon module, but you can leave it as it is and resolve this conversation

Kraemii (approved):

PR #569, opened 2021-12-20
Fix: missing category of notus advisories

Kraemii:
**What**:
Add category for notus advisories
SC-479

**Why**:
gvmd got error due to missing category

**How**:
Tested with the get vt command that the notus advisories contains the category now:
`gvm-cli --log DEBUG --timeout 3600 --protocol OSP socket --socketpath /home/ck2104/greenbone/install/var/run/ospd/ospd-openvas.sock -X "<get_vts/>" | xmlstarlet fo > all_vts.xml`
xmlstarlet is for formatting the output, to be able to read the data properly I saved them in a file.

**Checklist**:

- [x] Tests

### DIFF
--- a/ospd_openvas/notus.py
+++ b/ospd_openvas/notus.py
@@ -122,6 +122,7 @@ class Notus:
         result["refs"] = refs
         result["family"] = path.stem
         result["name"] = advisory.get("title", "")
+        result["category"] = "3"
         return result
 
     def get_filenames_and_oids(self):


bjoernricks (approved):
Do you know what's the meaning of `"3"`? Would love if we move this value into an self describing constant variable in another PR.

Kraemii:
I am not really sure, so please correct me if I am wrong. I think the category is normally related to the order in which scripts are executed. This would make sense with notus. The different definitions can be found within the openvas-scanner and are:
```C
typedef enum
{
  ACT_INIT = 0,
  ACT_SCANNER,
  ACT_SETTINGS,
  ACT_GATHER_INFO,
  ACT_ATTACK,
  ACT_MIXED_ATTACK,
  ACT_DESTRUCTIVE_ATTACK,
  ACT_DENIAL,
  ACT_KILL_HOST,
  ACT_FLOOD,
  ACT_END,
} nvt_category;
```

PR #581, opened 2022-01-13
Fix: hash calculation

nichtsfrei:
Due to yielding of notus as well as nvt oids the hash calculation failed due to incorrect order. To fix this the resulting list needs to be order by oids in lexicographical order.

To verify it run:
```
gvmd --rebuild
```

There should be no: Rebuilding all NVTs because of a hash value mismatch.

### DIFF
--- a/ospd_openvas/vthelper.py
+++ b/ospd_openvas/vthelper.py
@@ -188,6 +188,7 @@ class VtHelper:
                 # treat oids for dependency lookup as a dict
                 oids = dict(vt_collection)
 
+        vt_selection.sort()
         for vt_id in vt_selection:
             vt = self.get_single_vt(vt_id, oids)
             yield (vt_id, vt)


ospd_openvas/nvticache.py
```diff
@@ -91,13 +91,20 @@ def get_oids(self) -> Iterator[Tuple[str, str]]:
...
         if self.notus:
             for f, oid in self.notus.get_filenames_and_oids():
-                yield (f, oid)
+                result.append((f, oid))
         if self.ctx:
             for f, oid in OpenvasDB.get_filenames_and_oids(self.ctx):
                 if not self.notus or not self.notus.exists(oid):
-                    yield (f, oid)
+                    result.append((f, oid))
+        result.sort()
```
timopollmeier, line 106: I think you should sort the list by the oid element of the tuples like this:
```suggestion
        result.sort(key=lambda t: t[1])
```
Also, gvmd only needs the VTs to be sorted for the hash calculation, so it may be a better idea to create a list of just the OIDs in the hash calculation function and then sort this list. That way the list items do not have to store the file names, saving some memory.

nichtsfrei: right, my mistake it wasn't sorted by oid but by filename; for the second I remark as far as I can tell it is calling in this case `<get_vts filter='%s'/>` in the end so I don't see how it is limited by hash calculation?

nichtsfrei: Ahh I think now I got it. The order doesn't matter while processing the actual NVT in gvmd at all since it does it per oid anyway.

y0urself (approved):

PR #592, opened 2022-01-26
Fix: stacktrace on sensor when unexpectedly closing a ssh connection

nichtsfrei:
In the circumstances that ospd-openvas is run on a sensor and while
starting a scan a ssh connection got unexpectedly closed it logged a
stacktrace and assumingly crashed ospd-openvas.

Since this is self correcting occurance it prints a warning instead.


jjnicola (approved):
lgtm

PR #604, opened 2022-02-16
Fix: Stop and resume scan

jjnicola:
**What**:
Fix stop and resume task

Jira: AP-1893
Jira: SC-233

**Why**:
When a scan is started, stopped when it is still 0% progress and quickly resumed, the task ended as interrupted, or not in a blocking situation with "requested" status.

**How**:
1a) Start a vulnerability scan against a single host target
2a) As soon as the scan task has the status "0%", stop the scan task
3a) Once the scan task has the status "Stopped at 0%", resume the scan task
4a) If the scan task has the status "Stopped at 0%" or "Interrupted at 0%" subsequently, resume the scan task again

With the patch, it is possible to do this and the scan is stopped successfully and it is possible to resume it with out issues.

**Checklist**:

- [ ] Tests

### DIFF
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -973,9 +973,13 @@ class OSPDopenvas(OSPDaemon):
     @staticmethod
     def is_openvas_process_alive(openvas_process: psutil.Popen) -> bool:
 
-        if openvas_process.status() == psutil.STATUS_ZOMBIE:
-            logger.debug("Process is a Zombie, waiting for it to clean up")
-            openvas_process.wait()
+        try:
+            if openvas_process.status() == psutil.STATUS_ZOMBIE:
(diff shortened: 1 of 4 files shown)


ospd_openvas/daemon.py
```diff
@@ -1157,6 +1162,8 @@ def exec_scan(self, scan_id: str):
 
                 # clean main_db, but wait for scanner to finish.
                 while not kbdb.target_is_finished(scan_id):
+                    if not self.is_openvas_process_alive(openvas_process):
+                        break
```
Kraemii, line 1166:
```suggestion
                while not kbdb.target_is_finished(scan_id) and self.is_openvas_process_alive(openvas_process):
```

jjnicola: I want to break the while only if the new if statement is not true.

Kraemii (approved):

PR #618, opened 2022-03-21
Add: possibility to handle redis tcp connections configured in openvas

nichtsfrei:
Since https://github.com/greenbone/gvm-libs/pull/669 openvas can be
configured with tcp:// this commit enables the same behaviour within
ospd-openvas.

### DIFF
--- a/ospd_openvas/db.py
+++ b/ospd_openvas/db.py
@@ -23,6 +23,7 @@ import sys
 import time
 
 from typing import List, NewType, Optional, Iterable, Iterator, Tuple
+from urllib import parse
 
 import redis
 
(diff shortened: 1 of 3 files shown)


ospd_openvas/db.py
```diff
@@ -74,12 +75,24 @@ def get_database_address(cls) -> Optional[str]:
...
+                        "This feature is experimental and insecure. "
+                        "It is not recommended in production environments."
+                    )
 
         return cls._db_address
 
     @classmethod
     def create_context(
-        cls, dbnum: Optional[int] = 0, encoding: Optional[str] = 'latin-1'
+        cls, dbnum: Optional[int] = 0, encoding: str = 'latin-1'
```
Kraemii, line 95: Same for dbnum

PR #632, opened 2022-04-26
Change: Allow openvas access raw sockets within container

bjoernricks:
**What**:

Allow openvas access raw sockets within container

**Why**:

For the boreas alive detection it is required to access the raw socket. For capturing packages (for example used for detecting log4j) the NET_ADMIN capabilities are required.

**How**:

**Checklist**:

- [ ] Tests

### DIFF
--- a/.docker/prod.Dockerfile
+++ b/.docker/prod.Dockerfile
@@ -13,6 +13,7 @@ WORKDIR /ospd-openvas
 RUN apt-get update && \
     apt-get install --no-install-recommends --no-install-suggests -y \
     gosu \
+    libcap2-bin \
     python3 \
     python3-pip && \
     apt-get remove --purge --auto-remove -y && \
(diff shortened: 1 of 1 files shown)


.docker/prod.Dockerfile
```diff
@@ -30,6 +31,8 @@ RUN chgrp -R ospd-openvas /etc/openvas/ && \
 
 COPY dist/* /ospd-openvas
 
+# allow openvas to access raw sockets
+RUN setcap cap_net_raw=ep /usr/local/sbin/openvas
```
mawiegand, line 35: I would recommend using
```suggestion
RUN setcap cap_net_raw+ep /usr/local/sbin/openvas
```
to prevent overriding existing caps.

Dexus (approved):
As I already mentioned it in the chat. You have set the rights on `/usr/local/sbin/openvas` now correct.
I would recommend only to add a notice that the container needs a hostname, domainname and the container its self needs also both caps, also for log4j scans it should listen to its own IP which can be done with network mode: host, ipvlan, macvlan or other plugins (need to be able to listen to all ports tcp/udp)

PR #646, opened 2022-05-04
Fix: check for param to be present in calculate_vts_collection_hash

nichtsfrei:
To prevent crashs when a vt was ignored (e.g. when no severity was
provided) calculate_vts_collection_hash should verify if a param
before using it.


Kraemii (approved):

ospd_openvas/vthelper.py
```diff
@@ -210,11 +210,12 @@ def calculate_vts_collection_hash(self) -> str:
             vt_params = vt.get('vt_params')
```
Kraemii, line 210: Isn't this line causing the issue?

Kraemii (approved):

PR #647, opened 2022-05-05
Refactor: remove on_disonnect on favor of destructor

nichtsfrei:
In paho version 1.5.1 (used in debian) _do_on_disconnect can raise an
exception due to incorrect usage. Since we cannot change the behaviour
of the packaged version we switch from using on_disonnect to a
destructor approach.

jira: SC-599 SC-606


Kraemii:
With this patch we don't get anything logged when using INFO as Log-level

Kraemii (approved):

PR #649, opened 2022-05-08
[ospd-openvas.service] Drop Group= directive

Flowdalic:
Specifying the Group is typically never required, as systemd's default
behavior to use the user's primary group is what you want anyways. And
to make the service file more readable, unnecessary directives should
be removed. So this removes the Group= directive.


nichtsfrei:
verified via `man 5  systemd.exec`: `If no group is set, the default group of the user is used.`.

Thanks.

nichtsfrei (approved):

PR #680, opened 2022-06-14
notus smoketest

nichtsfrei:
For allowing Notus tests we fake a Notus server by using paho MQTT and
returning messages for openvas to finish the scan and ospd-openvas
addtend the results.

Afterwards we verify that report results contain the wanted message.

Can only be merged after https://github.com/greenbone/ospd-openvas/pull/678


smoketest/mqtt/mqtt.go
```diff
@@ -0,0 +1,135 @@
...
+
+func (m MQTT) Close() error {
+	close(m.in)
+	return m.client.Disconnect(&paho.Disconnect{ReasonCode: 0})
+}
+
+func (m MQTT) register(topic string) error {
+
+	m.client.Router.RegisterHandler(topic, func(p *paho.Publish) {
+		m.in <- &TopicData{Topic: topic, Message: p.Payload}
```
Kraemii, line 38: Maybe I am a bit confused, but I think the name `in` is very irritating, as it is used to receive the actual data send via mqtt (if I understand it correctly)

nichtsfrei: It is the data received not send from mqtt hence in for incoming.

Kraemii (approved):

PR #688, opened 2022-07-07
Refactor instead of setting filter to family load oids

nichtsfrei:
When loading a policy with all families ospd-scans should load all nvts
with a registered family instead of a filter.

In addition scans/main.go got extended to be capable of running multiple
policies and print the progress.

### DIFF
--- a/smoketest/cmd/scans/main.go
+++ b/smoketest/cmd/scans/main.go
@@ -1,12 +1,15 @@
 package main
 
 import (
+	"encoding/xml"
 	"flag"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"strings"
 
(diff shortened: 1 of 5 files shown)


ArnoStiefvater (changes requested):
Can you adjust the commit message. `Since gvmd does actually load of oids instead of sending filter to ospd the smoke test should do the dame.` is not really comprehensible.

Else looks good.

nichtsfrei:
> Can you adjust the commit message. `Since gvmd does actually load of oids instead of sending filter to ospd the smoke test should do the dame.` is not really comprehensible.

Changed the commit message :)

ArnoStiefvater (approved):

PR #693, opened 2022-07-14
Change: make get_status of scan more robust

nichtsfrei:
In some cases it can happen that a scan_id is not findable since it will
not block the execution of ospd-openvas it should not printed as an
error with a stacktrace but as a warning.


ArnoStiefvater (dismissed):
Blocking accidental merge because of external request. Should be merged after new major release is out.

Kraemii (dismissed):

ArnoStiefvater (approved):

PR #709, opened 2022-07-27
Change: when sha256sums for notus are missing ignore instead of crash

nichtsfrei:
This commit changes the behaviour of ospd-openvas quitting when
sha256sums files are missing to just ignoring the advisories and print a
WARNING like:
```
OSPD[316986] 2022-07-27 13:20:01,611: WARNING: (ospd_openvas.notus) ignoring /var/lib/notus/advisories/my test family.notus due to invalid signature
```

### DIFF
--- a/ospd_openvas/gpg_sha_verifier.py
+++ b/ospd_openvas/gpg_sha_verifier.py
@@ -44,13 +44,18 @@ def reload_sha256sums(
         # therefore a collision is not the end of the world and sha1 is more
         # than sufficient
         hasher = hashlib.sha1()
-        with file.open(mode="rb") as f:
-            for hash_file_bytes in iter(lambda: f.read(1024), b""):
-                hasher.update(hash_file_bytes)
-        return hasher.hexdigest()
+        try:
+            with file.open(mode="rb") as f:
+                for hash_file_bytes in iter(lambda: f.read(1024), b""):
+                    hasher.update(hash_file_bytes)
+            return hasher.hexdigest()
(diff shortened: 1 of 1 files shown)


ospd_openvas/gpg_sha_verifier.py
```diff
@@ -44,13 +44,18 @@ def create_hash(file: Path) -> str:
...
-        with file.open(mode="rb") as f:
-            for hash_file_bytes in iter(lambda: f.read(1024), b""):
-                hasher.update(hash_file_bytes)
-        return hasher.hexdigest()
+        try:
+            with file.open(mode="rb") as f:
+                for hash_file_bytes in iter(lambda: f.read(1024), b""):
+                    hasher.update(hash_file_bytes)
+            return hasher.hexdigest()
+        except:
```
bjoernricks, line 52: Maybe just catch a specific base error here? A catchall can hide serious issues.

bjoernricks: I guess [OSError](https://docs.python.org/3/library/exceptions.html#OSError) would be best here

bjoernricks (approved):
Looks good to me. I guess this should be backported to stable? Does ospd-openvas try to load the file periodically? or does it need a restart to load it when the hashfile is available finally?

nichtsfrei:
It should reload periodically; backport will be done later with the other bug fixes to code freeze within stable.

ArnoStiefvater (approved):

PR #710, opened 2022-07-28
Refactor: use defaults as in prod.Dockerfile for smoketests

nichtsfrei:


smoketest/cmd/scans/main.go
```diff
@@ -22,7 +22,7 @@ const protocoll = "unix"
 func init() {
 	address = os.Getenv("OSPD_SOCKET")
 	if address == "" {
-		address = "/var/run/ospd/ospd.sock"
+		address = "/var/run/ospd/ospd-openvas.sock"
```
bjoernricks, line 25:
```suggestion
		address = "/run/ospd/ospd-openvas.sock"
```

smoketest/cmd/test/main.go
```diff
@@ -25,7 +25,7 @@ const protocoll = "unix"
 func init() {
 	ospdSocket = os.Getenv("OSPD_SOCKET")
 	if ospdSocket == "" {
-		ospdSocket = "/var/run/ospd/ospd.sock"
+		ospdSocket = "/var/run/ospd/ospd-openvas.sock"
```
bjoernricks, line 28:
```suggestion
		ospdSocket = "/run/ospd/ospd-openvas.sock"
```

smoketest/run-tests.sh
```diff
@@ -15,15 +15,15 @@ mosquitto -c /etc/mosquitto.conf &
 redis-server /etc/redis/redis.conf
 /usr/sbin/sshd
 ospd-openvas --disable-notus-hashsum-verification True \
-  -u /var/run/ospd/ospd.sock \
+  -u /var/run/ospd/ospd-openvas.sock \
```
bjoernricks, line 18:
```suggestion
  -u /run/ospd/ospd-openvas.sock \
```

smoketest/run-tests.sh
```diff
@@ -15,15 +15,15 @@ mosquitto -c /etc/mosquitto.conf &
...
   -l /var/log/gvm/ospd.log
 wait_turn=0
-while [ ! -S /var/run/ospd/ospd.sock ]; do
+while [ ! -S /var/run/ospd/ospd-openvas.sock ]; do
```
bjoernricks, line 21:
```suggestion
while [ ! -S /run/ospd/ospd-openvas.sock ]; do
```

bjoernricks (commented):
`/var/run` is actually only a symlink to `/run`. Btw. the default path in https://github.com/greenbone/ospd-openvas/blob/main/ospd/parser.py#L35 should be changed too now since ospd is part of ospd-openvas.

smoketest/cmd/scans/main.go
nichtsfrei: I used a sed command to change every case instead of using suggestion; I hope you don't mind.

ArnoStiefvater (approved):

PR #713, opened 2022-08-02
Fix notus results

Kraemii:
**What**:
Automatically subscribes again to all topics after a reconnect of MQTT
SC-649

**Why**:
In case the ospd-openvas MQTT client disconnect from its broker, it dropped all settings for subscriptions, so MQTT was not working again after a reconnect. This issue is fixed with this PR

**How**:
Run ospd-openvas and notus-scanner. They should connect to a MQTT broker. After ospd-openvas is initialized stop the broker and restart it after a few seconds. Start a scan and watch for notus results. In the old version ospd-openvas should not be able to receive any. This PR fixes this issue.
Start scan via gvm-cli:
```xml
<start_scan scan_id='97079ee9-8917-49da-aa4f-4ef95f757ac1'>
    <vt_selection>
        <vt_single id='1.3.6.1.4.1.25623.1.0.50282'>
        </vt_single>
    </vt_selection>
    <targets>
        <target>
            <hosts>localhost</hosts>
            <ports>80,8080,443,22</ports>
            <credentials>
                <credential service="ssh" type="up" port="22">
                    <username>ssh_user</username>
                    <password>ssh_pw</password>
                </credential>
            </credentials>
        </target>
    </targets>
    <scanner_params>
    </scanner_params>
</start_scan>
```

**Checklist**:

- [ ] Tests
- [ ] PR merge commit message adjusted

### DIFF
--- a/ospd_openvas/messaging/mqtt.py
+++ b/ospd_openvas/messaging/mqtt.py
@@ -86,7 +86,13 @@ class MQTTPublisher(Publisher):
 
 class MQTTSubscriber(Subscriber):
     def __init__(self, client: MQTTClient):
-        self._client = client
+        self.client = client
+        # Save the active subscriptions on subscribe() so we can resubscribe
+        # after reconnect
+        self.subscriptions: dict = {}
+
+        self.client.on_connect = self.on_connect
+        self.client.user_data_set(self.subscriptions)
 
     def subscribe(
         self, message_class: Type[Message], callback: Callable[[Message], None]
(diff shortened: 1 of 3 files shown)


bjoernricks (changes requested):
Please support newer versions of paho-mqtt too just use >=1.5.1

ospd_openvas/messaging/mqtt.py
```diff
@@ -96,8 +101,22 @@ def subscribe(
 
         logger.debug("Subscribing to topic %s", message_class.topic)
 
-        self._client.subscribe(message_class.topic, qos=QOS_AT_LEAST_ONCE)
-        self._client.message_callback_add(message_class.topic, func)
+        self.client.subscribe(message_class.topic, qos=QOS_AT_LEAST_ONCE)
+        self.client.message_callback_add(message_class.topic, func)
+
+        self.subscriptions[message_class.topic] = func
+        self.client.user_data_set(self.subscriptions)
```
Kraemii, line 108: you do not have to set the userdata for each subscription, setting it once in the init method is sufficient

ArnoStiefvater:
> Please support newer versions of paho-mqtt too just use >=1.5.1

Currently the version 1.5.1 is in debian bullseye which has some serious bugs we have to work around. The on_disconnect function is not working in 1.5.1 resulting in a crash when the broker disconnects. We want to be aware of this while developing so we do not run into such problems again.

ArnoStiefvater (approved):

Issue #725, opened 2022-08-09
OpenVAS does not clear stale pid files when current pid matches old pid


Ajedi32:
### Expected behavior

When starting ospd-openvas in Docker after a hard shutdown, OpenVAS recognizes that the PID file at `/run/ospd/ospd-openvas.pid` is stale, removes it, and continues to start.

### Actual behavior

OpenVAS displays the following error:

```
OSPD[1] 2022-08-09 15:25:51,151: ERROR: (ospd.misc) There is an already running process. See /run/ospd/ospd-openvas.pid.
```

### Steps to reproduce

1. Follow the instructions at https://greenbone.github.io/docs/latest/22.4/container/index.html to set up the Greenbone Community Containers.
2. Kill the ospd-openvas suddenly with `docker compose <options> kill ospd-openvas`
3. Attempt to restart the container with `docker compose <options> up ospd-openvas`

### GVM versions

**gsa:** 22.04.0

**gvm:** 22.4.0~dev1

**openvas-scanner:** 22.4.1~dev1

**gvm-libs:** 22.4.1~dev1

### Environment

**Operating system:**

```
$ uname -a && cat /etc/lsb-release 
Linux PC2287 5.10.16.3-microsoft-standard-WSL2 #1 SMP Fri Apr 2 22:23:49 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
DISTRIB_ID=Ubuntu
DISTRIB_RELEASE=20.04
DISTRIB_CODENAME=focal
DISTRIB_DESCRIPTION="Ubuntu 20.04.3 LTS"
```

**Installation method / source:** Official docker-compose.yml file on https://greenbone.github.io/docs/latest/22.4/container/index.html

### Proposed fix

There was already an attempt to fix this in 200079a7a7857c3c267f93554cf7cd2811ec6b31. The fix doesn't work in a docker environment because with docker OpenVAS runs in a pid namespace where it always sees its pid as 1. Therefore the check at https://github.com/greenbone/ospd-openvas/blob/c10b8c266034d569267618097951dab1a31317d8/ospd/misc.py#L121 yields a false positive.

That line should be updated to also check if the pid in the file matches the pid of the current process, and clear the file if it does.

nichtsfrei:
Thank you for finding and reporting this bug.

Unfortunately it is not enough to verify if the PID used in the file is the same as used since they're both 1 as well.

[...]

PR #728, opened 2022-08-10
Fix: unable to restart container after it was killed

nichtsfrei:
When the process name found by the PID file and the PID in the PID file
are the same it should be treated as it is from another process so that
ospd-openvas can be started even after a sudden kill of a container.

To verify this change you can start ospd-openvasby executing:
```
docker run -v /run/redis/redis.sock:/run/redis/redis.sock \
  --name "narf" \
  --privileged greenbone/ospd-openvas:unstable ospd-openvas -f
```

Then kill it via:

```
docker kill narf
```

and restart it with

```
docker start -a narf
```

Fixes: https://github.com/greenbone/ospd-openvas/issues/725

### DIFF
--- a/ospd/misc.py
+++ b/ospd/misc.py
@@ -103,22 +103,23 @@ def create_pid(pidfile: str) -> bool:
     Otherwise gives an error."""
 
     pid = str(os.getpid())
-    new_process = psutil.Process(int(pid))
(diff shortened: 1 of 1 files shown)


Ajedi32:
👍 LGTM

Kraemii (approved):

PR #750, opened 2022-08-23
Fix: Stop and clean finished/running openvas process before resumming a scan

jjnicola:
**What**:
Fix: Stop and clean finished/running openvas process before resumming a scan

Jira: SC-624

**Why**:
If Ospd-Openvas crashes during a scan, probably the scan will continue
running in background, leaving the kb in redis occupied.
After ospd-openvas is initialized, the task can be resumed.
In this case, the scan id of the task will be the same one as before,
and there will be two kbs for the old and the new tasks with the same scanid.

This patch ensure that there is no kb in redis in used with the same scan id,
and also it will stop the scan in case it is still running.

**How**:
- Run a scan and kill ospd-openvas with SIGKILL (-9) signal: `killall -9 ospd-openvas`
- Start ospd-openvas 
- Check with top/htop/ps, that openvas runs in background, it is an orphan process with ppid = 1 (system).
- Check in gvmd that the task is stopped or interrupted
- Check in redis that there is a main kb in use and the scan id. If you started the test with a clean setup, it should be the kb 2
- Resume the task. It will use the same scan id. Compare it whit the one checked in redis.

- Without the patch: the old openvas process still run and a new kbs are taken with the same scan id for the resume task
- With the patch, the old openvas process are stopped and its kb is released before resuming the scan.

**Checklist**:

- [ ] Tests

### DIFF
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -1090,6 +1090,15 @@ class OSPDopenvas(OSPDaemon):
             dryrun.exec_dry_run_scan(scan_id, self.nvti, OSPD_PARAMS)
             return
 
+        kbdb, err = self.main_db.check_consistency(scan_id)
+        if err < 0:
+            logger.debug(
+                "An old scan with the same scanID was found in the kb. "
+                "Waiting for the kb clean up to finish."
+            )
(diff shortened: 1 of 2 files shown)


jjnicola:
Close in favor of #751

PR #751, opened 2022-08-23
Fix: Stop and clean finished/running openvas process before resumming a scan

jjnicola:
**What**:
Fix: Stop and clean finished/running openvas process before resumming a scan

Jira: SC-624

**Why**:
If Ospd-Openvas crashes during a scan, probably the scan will continue
running in background, leaving the kb in redis occupied.
After ospd-openvas is initialized, the task can be resumed.
In this case, the scan id of the task will be the same one as before,
and there will be two kbs for the old and the new tasks with the same scanid.

This patch ensure that there is no kb in redis in used with the same scan id,
and also it will stop the scan in case it is still running.

**How**:
- Run a scan and kill ospd-openvas with SIGKILL (-9) signal: `killall -9 ospd-openvas`
- Start ospd-openvas 
- Check with top/htop/ps, that openvas runs in background, it is an orphan process with ppid = 1 (system).
- Check in gvmd that the task is stopped or interrupted
- Check in redis that there is a main kb in use and the scan id. If you started the test with a clean setup, it should be the kb 2
- Resume the task. It will use the same scan id. Compare it whit the one checked in redis.

- Without the patch: the old openvas process still run and a new kbs are taken with the same scan id for the resume task
- With the patch, the old openvas process are stopped and its kb is released before resuming the scan.

**Checklist**:

- [ ] Tests

### DIFF
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -1180,7 +1180,9 @@ class OSPDopenvas(OSPDaemon):
         while kbdb.get_status(scan_id) == 'new':
             res = openvas_process.poll()
             if res and res < 0:
-                self.stop_scan_cleanup(kbdb, scan_id, openvas_process.pid)
+                self.stop_scan_cleanup(
+                    kbdb, scan_id, kbdb.get_scan_process_id()
+                )
                 logger.error(
                     'It was not possible run the task %s, since openvas ended '
                     'unexpectedly with errors during launching.',
@@ -1207,7 +1209,9 @@ class OSPDopenvas(OSPDaemon):
             if scan_stopped:
                 logger.debug('%s: Scan stopped by the client', scan_id)
 
-                self.stop_scan_cleanup(kbdb, scan_id, openvas_process.pid)
+                self.stop_scan_cleanup(
+                    kbdb, scan_id, kbdb.get_scan_process_id()
+                )
 
                 # clean main_db, but wait for scanner to finish.
                 while not kbdb.target_is_finished(scan_id):
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -1090,6 +1090,15 @@ class OSPDopenvas(OSPDaemon):
             dryrun.exec_dry_run_scan(scan_id, self.nvti, OSPD_PARAMS)
             return
 
+        kbdb, err = self.main_db.check_consistency(scan_id)
+        if err < 0:
+            logger.debug(
+                "An old scan with the same scanID was found in the kb. "
+                "Waiting for the kb clean up to finish."
+            )
+            self.stop_scan_cleanup(kbdb, scan_id, kbdb.get_scan_process_id())
+            self.main_db.release_database(kbdb)
+
         do_not_launch = False
         kbdb = self.main_db.get_new_kb_database()
         scan_prefs = PreferenceHandler(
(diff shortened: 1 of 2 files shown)


Kraemii:
With this PR I am not able to stop a scan, ospd-openvas states that the scan stopped the scan, but openvas continues running. The status of the scan in gsa stays on stop requested. The ospd-openvas process, which handles the scan gets stucked. It also seems that no stop scan signal is sent to openvas, as when I manually send a SIGUSR1, the scan will stop and everything seems working again

ospd_openvas/daemon.py
```diff
@@ -1031,8 +1031,17 @@ def stop_scan_cleanup(
             # Set stop flag in redis
             kbdb.stop_scan(scan_id)
 
+            try:
+                ovas_process = psutil.Process(int(ovas_pid))
+            except psutil.NoSuchProcess:
+                ovas_process = None
```
Kraemii, line 1037: In case openvas is not running anymore this PID can be belong to any running process which happens to get the same PID as the old openvas process. The processes from psutil on the other hand will check if this process is the same, so it will guarantee that it is not performing actions on another process, which happens to have the same PID

jjnicola: Yes, therefore I check for the Process name, to ensure that it is an "openvas" process

jjnicola: I changed the signature to use the PID and not the Process, because in the case I am trying to handle with this PR, the orignal process doesn't exist (it is a new process with other PID), but what I want to terminate is the old one, still runnning. I have for that old scan just the PID.

Kraemii: Ah you are right

jjnicola:
> With this PR I am not able to stop a scan, ospd-openvas states that the scan stopped the scan, but openvas continues running. The status of the scan in gsa stays on stop requested. The ospd-openvas process, which handles the scan gets stucked. It also seems that no stop scan signal is sent to openvas, as when I manually send a SIGUSR1, the scan will stop and everything seems working again

Well, I found the issue. The problem is actually the name process check, since for the running scan (normal running), the PID belogns to the psutil.Popen, which has the name "sudo" while I was checking for "openvas". Therefore the the scan was never stop, as it was not considered an openvas process. Now, for stopping a normal scan, I send the PID got it from the kb.

ospd_openvas/daemon.py
jjnicola: Fixed with 137f70e9793cf17d26b16c0102847d47423d8d61

ospd_openvas/daemon.py
```diff
@@ -1081,6 +1090,15 @@ def exec_scan(self, scan_id: str):
...
             return
 
+        kbdb, err = self.main_db.check_consistency(scan_id)
+        if err < 0:
+            logger.debug(
+                "An old scan with the same scanID was found in the kb. "
+                "Waiting for the kb clean up to finish."
+            )
+            self.stop_scan_cleanup(kbdb, scan_id, kbdb.get_scan_process_id())
+            self.main_db.release_database(kbdb)
```
Kraemii, line 1100:
```suggestion
```
database is already cleaned in `stop_scan_cleanup()`

jjnicola: Do you mean this lines at end of `stop_scan_cleanup()`?
``` python
 # Clean redis db
            for scan_db in kbdb.get_scan_databases():
                self.main_db.release_database(scan_db)
```
This lines clean the "host's kbs", but not the main kb.
We do the same below, e.g [here](https://github.com/greenbone/ospd-openvas/blob/0772e489128d01f4102b7bdc112c28b3a08ca026/ospd_openvas/daemon.py#L1225)

Kraemii (approved):

Issue #765, opened 2022-09-06
[22.4.2] - raise Exception("GPG verification of notus sha256sums failed")


tgurr:
Running into an error with the latest version 22.4.2 (downgrading back to 22.4.0 resolves the problem).

### Expected behavior

Starting up and running without any issues (like 22.4.0 did and still does for me).

### Actual behavior

Running into the following error since upgrading from 22.4.0 to 22.4.2.

```
Sep 06 18:34:57 hostname ospd-openvas[4407]: Traceback (most recent call last):
Sep 06 18:34:57 hostname ospd-openvas[4407]:   File "/usr/host/bin/ospd-openvas", line 33, in <module>
Sep 06 18:34:57 hostname ospd-openvas[4407]:     sys.exit(load_entry_point('ospd-openvas==22.4.2', 'console_scripts', 'ospd-openvas')())
Sep 06 18:34:57 hostname ospd-openvas[4407]:   File "/usr/lib/python3.10/site-packages/ospd_openvas/daemon.py", line 1243, in main
Sep 06 18:34:57 hostname ospd-openvas[4407]:     daemon_main('OSPD - openvas', OSPDopenvas, NotusParser())
Sep 06 18:34:57 hostname ospd-openvas[4407]:   File "/usr/lib/python3.10/site-packages/ospd/main.py", line 164, in main
Sep 06 18:34:57 hostname ospd-openvas[4407]:     daemon.init(server)
Sep 06 18:34:57 hostname ospd-openvas[4407]:   File "/usr/lib/python3.10/site-packages/ospd_openvas/daemon.py", line 524, in init
Sep 06 18:34:57 hostname ospd-openvas[4407]:     self.update_vts()
Sep 06 18:34:57 hostname ospd-openvas[4407]:   File "/usr/lib/python3.10/site-packages/ospd_openvas/daemon.py", line 649, in update_vts
Sep 06 18:34:57 hostname ospd-openvas[4407]:     self.nvti.notus.reload_cache()
Sep 06 18:34:57 hostname ospd-openvas[4407]:   File "/usr/lib/python3.10/site-packages/ospd_openvas/notus.py", line 119, in reload_cache
Sep 06 18:34:57 hostname ospd-openvas[4407]:     if self._verifier(f):
Sep 06 18:34:57 hostname ospd-openvas[4407]:   File "/usr/lib/python3.10/site-packages/ospd_openvas/gpg_sha_verifier.py", line 121, in verify
[...]

PR #767, opened 2022-09-08
Change: gpg verification failure handling of notus advisories sha256sums

nichtsfrei:
When gpg verification on sha256sums for notus advisories fails it is
printing a warning instead of crashing ospd-openvas.

This changes the behaviour mentioned in https://github.com/greenbone/ospd-openvas/issues/765


nichtsfrei:
Already merged with another PR.

PR #768, opened 2022-09-08
Fix: use ~/.gnupg when /etc/openvas/gnupg is not available

nichtsfrei:
Instead of just using GOS defaults `/etc/openvas/gnupg` check if the
directory exists and when not use `$HOME/.gnupg` instead.

If both are not available print a warning that the env variable
GNUPGHOME should be set but stick with the failing `$HOME/.gnupg` to
prevent None checking.

Fixes https://github.com/greenbone/ospd-openvas/issues/765

Depends on: https://github.com/greenbone/ospd-openvas/pull/767

### DIFF
--- a/ospd_openvas/gpg_sha_verifier.py
+++ b/ospd_openvas/gpg_sha_verifier.py
@@ -1,12 +1,27 @@
 import hashlib
 import os
+import logging
 from pathlib import Path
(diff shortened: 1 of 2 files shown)


y0urself (approved):
Looks very good to me! :)

PR #773, opened 2022-09-21
Separate notus redis cache from openvas cache

jjnicola:
**What**:
Separate notus redis cache from openvas cache.
Depends on PR #772 
Jira: SC-692

**Why**:
It makes caches independent of each other. Now each cache has it owns methods to handle the oids. Before, the nvticache was initialized with the notus cache. There was only one get_oids methods in the nvticache, which returned the notus advisories as well.
VtHelper is now initialized with both cache (Notus is still optional), and the oids can be get separately from one or another cache.

**How**:
Run the get_vts command. You should get the same amount of vts, in the same time

`time gvm-cli --protocol OSP --timeout 320  socket --sockpath /home/jnicola/install/var/run/ospd/openvas.sock --xml "<get_vts/>" |xmlstarlet fo |grep "<vt id" | wc -l` 

Also, the sha256:hash must the same for the same feed

`time gvm-cli --protocol OSP --timeout 320  socket --sockpath /home/jnicola/install/var/run/ospd/openvas.sock --xml "<get_vts version_only='1'/>" |xmlstarlet fo`

**Checklist**:

- [ ] Tests

### DIFF
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -385,10 +385,11 @@ class OpenVasVtsFilter(VtsFilter):
 
     """Methods to overwrite the ones in the original class."""
 
-    def __init__(self, nvticache: NVTICache) -> None:
+    def __init__(self, nvticache: NVTICache, notuscache: Cache) -> None:
         super().__init__()
 
         self.nvti = nvticache
+        self.notus = notuscache
 
     def format_vt_modification_time(self, value: str) -> str:
         """Convert the string seconds since epoch into a 19 character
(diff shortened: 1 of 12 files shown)


jjnicola:
Depends on #772

cfi-gb:
There is a `Sepatate` typo in the PR title, suggestion:

> Separate Notus Redis cache

nichtsfrei (approved):

ospd_openvas/daemon.py
```diff
@@ -1094,7 +1099,7 @@ def exec_scan(self, scan_id: str):
         do_not_launch = False
         kbdb = self.main_db.get_new_kb_database()
         scan_prefs = PreferenceHandler(
-            scan_id, kbdb, self.scan_collection, self.nvti
+            scan_id, kbdb, self.scan_collection, self.nvti, self.notus.exists
```
Kraemii, line 1102: `self.notus` might be `None`

jjnicola: We are passing a method to check if an oid is in the notus cache (not the notus object). 
Unfortunately, gvmd does not use the group filter for sending entire families of plugins (group:"Ubuntu Local Security Check), but it is passing each single OID in the plugins list.
This means that we need to check if the OID belongs to the notus cache or not, to excluded it. So, it is not sent to openvas as single plugin. The method notus.exists(OID) checks that.

LSC are run for all plugins. IMHO the clients should send the family name instead of list of OIDs. In case of receiving a the family name for a LSC, we could remove all notus stuff from the preferencehandler. But it is not the case, and I wouldn't expect such a change in the gvmd side.

Check the need and usage of the method [here](https://github.com/greenbone/ospd-openvas/blob/075bd851fc8dc0488d35364929f8f03c61a307a1/ospd_openvas/preferencehandler.py#L110) and [here](https://github.com/greenbone/ospd-openvas/blob/075bd851fc8dc0488d35364929f8f03c61a307a1/ospd_openvas/preferencehandler.py#L251)

PR #784, opened 2022-10-07
Add: Make openvas-scanner option visible for the client

jjnicola:
**What**:
Make openvas-scanner option visible for the client
Now, the client can see the option "test_alive_wait_timeout"

Jira: SC-680

**Why**:
Openvas has this option available, and it is desirable to set it from the client side.

**How**:
Send the

**Checklist**:

- [ ] Tests


jjnicola:
Although it is not a breaking change, it will only working after merging the dependent PR greenbone/gvm-libs#711

nichtsfrei (approved):

PR #787, opened 2022-10-10
Make the alive_test_ports scanner preference visible for the clients.

jjnicola:
**What**:
Make the alive_test_ports scanner preference visible for the clients.
The option can be sent to ospd-openvas as scanner preference now.

Jira: SC-689

**Why**:
Improve the alive detection. Now the user are able to specify a port list for the TCP SYN and TCP ACK alive test methods.
This scanner preference was before scanner preference only and hide for the clients.

**How**:
Configure a target for testing. E.g enter the target and set an iptable rule, so the port 35000 does not answer.
`sudo iptables -A INPUT -p tcp --dport 35000 -s 192.168.x.x -j DROP`
- Configure a target in GSA with the alive test method "TCP ACK" 

Run a scan and check that the host is found **_alive_** (maybe ssh port 22 is open)

- Create a new scan config and set the scanner preference "alive_test_ports = 35000"
- Run a scan against the target with the new scan config. The host should be find **_dead._**
 

**Checklist**:

- [ ] Tests

### DIFF
--- a/ospd_openvas/daemon.py
+++ b/ospd_openvas/daemon.py
@@ -295,6 +295,15 @@ OSPD_PARAMS = {
             + '"T:1-65535,U:1-1024".'
         ),
     },
+    'alive_test_ports': {
+        'type': 'string',
+        'name': 'alive_test_ports',
+        'default': '21-23,25,53,80,110-111,135,139,143,443,445,'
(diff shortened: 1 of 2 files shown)


jjnicola:
Depends on greenbone/gvm-libs#713

nichtsfrei (approved):

PR #788, opened 2022-10-11
Change: remove codecov

nichtsfrei:
Due to stability reason when uploading we decided to remove codecov.


y0urself (changes requested):
You need to remove codecov from REAMDE.

ArnoStiefvater (approved):

y0urself (approved):

PR #808, opened 2022-11-23
Add: ospd command to dump the string used for VTs verification

jjnicola:
**What**:
Add: ospd command to dump the string used for VTs verification

Jira: SC-739

**Why**:
This is for debugging purposes.

**How**:
```
$ gvm-cli --protocol OSP socket --socketpath /home/jnicola/install/var/run/ospd/openvas.sock --xml "<dump_vt_verification path='/tmp/jj.dump'/>" |xmlstarlet fo
<?xml version="1.0"?>
<dump_vt_verification_response status="200" status_text="OK">
  <hash>91e7ffe2c850bf9ce561529d8280a8e5af161c9f6f13d0212e2741029e224c45</hash>
</dump_vt_verification_response>
```

**Checklist**:

- [ ] Tests


ospd_openvas/vthelper.py
```diff
@@ -209,10 +210,15 @@ def get_vt_iterator(
             if vt:
                 yield (vt_id, vt)
 
-    def calculate_vts_collection_hash(self) -> str:
+    def calculate_vts_collection_hash(self, path: Path = None) -> str:
```
nichtsfrei, line 213: wouldn't it be easier and less error prone to use when the string to calculate it is returned when wanted instead of writing it into a file?

nichtsfrei (changes requested):

ospd_openvas/vthelper.py
jjnicola: the ospd command returns an xml object to the gmv-cli. If you want to work with the string, you have first to process the xml element. E.g. in this case, you can just apply the `sha256sum <file.dump>` to get the hash and compare it.

nichtsfrei: But in this case you make have to make sure that the path exists, is writable, ... and return an appropriate error code if not. So I think that dealing with xml is in comparison easier.

PR #843, opened 2023-01-26
Fix: ospd-openvas depends on mosquitto MQTT broker

bjoernricks:
**What**:

Update the ospd-openvas.service file to require a running mosquitto service.

**Why**:

Without mosquitto ospd-openvas will issue errors in the log

### DIFF
--- a/config/ospd-openvas.service
+++ b/config/ospd-openvas.service
@@ -2,7 +2,7 @@
 Description=OSPd Wrapper for the OpenVAS Scanner (ospd-openvas)
 Documentation=man:ospd-openvas(8) man:openvas(8)
 After=network.target networking.service redis-server@openvas.service mosquitto.service
-Wants=redis-server@openvas.service mosquitto.service
+Wants=redis-server@openvas.service mosquitto.service notus-scanner.service
 ConditionKernelCommandLine=!recovery
 
 [Service]
--- a/config/ospd-openvas.service
+++ b/config/ospd-openvas.service
@@ -1,8 +1,8 @@
 [Unit]
 Description=OSPd Wrapper for the OpenVAS Scanner (ospd-openvas)
 Documentation=man:ospd-openvas(8) man:openvas(8)
-After=network.target networking.service redis-server@openvas.service
-Wants=redis-server@openvas.service
+After=network.target networking.service redis-server@openvas.service mosquitto.service
+Wants=redis-server@openvas.service mosquitto.service
 ConditionKernelCommandLine=!recovery
 
 [Service]


cfi-gb:
Wouldn't be better to "fix" the message in ospd-openvas instead? There might be reasons to not use Notus scanner (not scanning Linux systems, too complex setup required, ...).

Or is this still possible after this addition?

bjoernricks:
This is about not being able to contact the mosquitto broker. ospd-openvas depends on the mosquitto broker to work correctly at the moment. Therefore this fix is required. 

Currently there is no possibility (as far as I know) to deactivate the notus-scanner integration with some setting. If you think that is a problem you might talk to the scanner team and discuss this. But for me it is out of scope of this PR.

cfi-gb:
So the following:

> Configures (weak) requirement dependencies on other units. This option may be specified more than once or multiple space-separated units may be specified in one option in which case dependencies for all listed names will be created. Dependencies of this type may also be configured outside of the unit configuration file by adding a symlink to a .wants/ directory accompanying the unit file. For details, see above.
>
> Units listed in this option will be started if the configuring unit is. However, if the listed units fail to start or cannot be added to the transaction, this has no impact on the validity of the transaction as a whole, and this unit will still be started. This is the recommended way to hook the start-up of one unit to the start-up of another unit.

from https://www.freedesktop.org/software/systemd/man/systemd.unit.html answers my question here:

> Or is this still possible after this addition?

In short:

`ospd-openvas.service` will still start if `mosquitto.service` isn't installed so it improves the situation for these having it while keeping it working for those who don't.

So this change should be absolutely valid :+1:

Additions:
- We could also add a `Wants=` for `notus-scanner.service` because that should be also started once `ospd-openvas.service` is started. Will add an in-line comment accordingly.
- Reference to the related community docs PR: greenbone/docs#258
- On a related note: It seems it is possible to disable `notus-scanner` integration by setting `mqtt_broker_address` to an empty string: https://github.com/greenbone/ospd-openvas/blob/5b25f14f4fd05338ea9ef503b398a926caa64974/ospd_openvas/daemon.py#L534-L538

config/ospd-openvas.service
```diff
@@ -1,8 +1,8 @@
 [Unit]
 Description=OSPd Wrapper for the OpenVAS Scanner (ospd-openvas)
 Documentation=man:ospd-openvas(8) man:openvas(8)
-After=network.target networking.service redis-server@openvas.service
-Wants=redis-server@openvas.service
+After=network.target networking.service redis-server@openvas.service mosquitto.service
+Wants=redis-server@openvas.service mosquitto.service
```
cfi-gb, line 5:
```suggestion
Wants=redis-server@openvas.service mosquitto.service notus-scanner.service
```

- As a default setup of `ospd-openvas` expects the `notus-scanner` integration it probably could make sense to include this as well (`notus-scanner.service` is also pulling in `mosquitto.service`)
- Not adding it to `After` as AFACT the service could be started at any time (while `mosquitto.service` needs to be actually started before `ospd-openvas.service`

nichtsfrei:
Those are warnings not errors and as Christian pointed out: mosquitto is just needed for a subfunctionality and is not a hard dependency. Therefore a warning message is correct; the frequency should be adapted though.

However since this is just a example systemd configuration it is also correct that for a full functionality mosquitto is unfortunately required.

nichtsfrei (approved):

bjoernricks:
Just a side note to make all aware of the importance of our defaults in the service and config file _"examples"_, Kali (and therefore Debian and all its derivatives) are [just patching our files](https://salsa.debian.org/pkg-security-team/ospd-openvas/-/blob/debian/master/debian/patches/Adapt-ospd-openvas.service.patch) and shipping them.

nichtsfrei:
I didn't mean to downplay the role of examples. It is just meant to be for the default and not exceptions (e.g. a runtime environment without Notus).

cfi-gb:
It should be noted that these are still examples and might not fit for every Linux distributions. Furthermore there might be various pitfalls in systemd related configurations a development team (like the scanner team providing these) isn't aware of. Usually package maintainers have more throughout knowledge on their currently maintained environments and systemd in general.

bjoernricks:
> Usually package maintainers have (hopefully) more throughout knowledge on their currently maintained environments and systemd in general.

I thought and expected that too but in reality that's not the case. Most likely the packagers are just following the defaults provided by upstream (us) because they are maintaining a lot of different packages. My initial intention was also these files are just examples but it turned out they are actually used without many changes. Therefore we need to provide this files with sane and usable defaults.

cfi-gb:
Just wanted to add this as an edit but will add it as a separate comment:

Yes, indeed. In the end it probably requires a cooperation with the related dev teams (to give / incorporate their knowledge of the component like which service needs to be started before another) and the package maintainer (to incorporate their knowledge of their current supported environment) to keep these examples in a state working for most standard systems.

Maybe some kind of disclaimer/hint in the service file could help?

PR #865, opened 2023-02-17
Change: call greenbone-feed-sync

jjnicola:
## What
Change: call greenbone-feed-sync
Depends on greenbone/greenbone-feed-sync#28

Jira: SC-773

## Why
Since the greenbone-nvt-sync will be removed, and replace with the new one greenbone-feed-sync

## Checklist

- [ ] Tests

### DIFF
--- a/ospd_openvas/feed.py
+++ b/ospd_openvas/feed.py
@@ -29,7 +29,7 @@ class Feed:
         """Calls greenbone-nvt-sync with --selftest option"""
 
         try:
-            return subprocess.getstatusoutput("greenbone-nvt-sync --selftest")
+            return subprocess.getstatusoutput("greenbone-feed-sync --selftest")
         except (subprocess.SubprocessError, OSError) as e:
             # the command is not available
             logger.warning('Feed sync self test failed. Reason %s', e)
--- a/ospd_openvas/feed.py
+++ b/ospd_openvas/feed.py
@@ -26,7 +26,7 @@ class Feed:
     """Class for getting feed info and checks"""
 
     def perform_feed_sync_self_test_success(self) -> tuple:
-        """Calls greenbone-nvt-sync with --selftest option"""
+        """Calls greenbone-feed-sync with --selftest option"""
 
         try:
             return subprocess.getstatusoutput("greenbone-feed-sync --selftest")
--- a/README.md
+++ b/README.md
@@ -66,6 +66,9 @@ file too:
 
     Defaults        secure_path=<existing paths...>:<install prefix>/sbin
 
+`ospd-openvas` calls an external tool to test if a feed update is possible.
+For this, the [greenbone-feed-sync](https://github.com/greenbone/greenbone-feed-sync) external tool is required.
+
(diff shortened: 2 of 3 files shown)


ospd_openvas/feed.py
```diff
@@ -29,7 +29,7 @@ def perform_feed_sync_self_test_success(self) -> tuple:
         """Calls greenbone-nvt-sync with --selftest option"""
```
bjoernricks, line 29:
```suggestion
        """Calls greenbone-feed-sync with --selftest option"""
```

bjoernricks (commented):
Personally I have doubts about the feed test concept here and why it is necessary in ospd-openvas. What's the purpose?

Currently it is an implicit dependency on an external tool. This needs to be documented in the README and/or installation documentation. Alternatively it would even better to make the dependency explicit and add a Python dependency on greenbone-feed-sync. This would also allow to call a Python API on greenbone-feed-sync instead of running a subprocess.

jjnicola:
> Personally I have doubts about the feed test concept here and why it is necessary in ospd-openvas. What's the purpose?

There is an OSP command which gets this information. This was required sometime ago.
``` xml
<check_feed_response status="200" status_text="OK">
  <feed>
    <lockfile_in_use>0</lockfile_in_use>
    <self_test_exit_error>0</self_test_exit_error>
    <self_test_error_msg/>
  </feed>
</check_feed_response>
```
 Also, there is a GMP command showing the error why a sync is not possible to be done.

```
<get_feeds_response status="200" status_text="OK">
  <feed>
    <type>NVT</type>
    <name>Greenbone Community Feed</name>
    <version>202302161010</version>
    <description>This script synchronizes an NVT collection with the 'Greenbone Community Feed'.
The 'Greenbone Community Feed' is provided by 'Greenbone Networks GmbH'.
Online information about this feed: 'https://www.greenbone.net/en/feed-comparison/'.
</description>
    <sync_not_available>
      <error>The sha256sum binary could not be found.</error>
    </sync_not_available>
  </feed>
---------------
```

It can be discussed if it is really necessary, and given the case, this can be removed from ospd-openvas, gvmd and gsa (in case it is used), and from the OSP and GMP protocols as well.

jjnicola:
> Currently it is an implicit dependency on an external tool. This needs to be documented in the README and/or installation documentation. Alternatively it would even better to make the dependency explicit and add a Python dependency on greenbone-feed-sync. This would also allow to call a Python API on greenbone-feed-sync instead of running a subprocess.

I prefer to avoid adding new dependencies. I have update the README.md file mentioning the script as an optional requirement.

bjoernricks:
> I prefer to avoid adding new dependencies. I have updated the README.md file mentioning the script as an optional requirement.

Just to make that clear: Actually you have this dependency. It's just implicit and not making it explicit isn't any better. The only difference is that you can install greenbone-feed-sync after ospd-openvas when using pip currently.

jjnicola:
Closing in favor of PR #896

PR #868, opened 2023-02-24
Add: option to use nasl-cli instead of openvas for feed update

nichtsfrei:
Adds a new option `--feed-updater` to switch between `openvas -u` and
`nasl-cli feed update`.

With this ospd-openvas can use the newly created `nasl-cli` to improve
the speed of importing nasl feed.

### DIFF
--- a/ospd/parser.py
+++ b/ospd/parser.py
@@ -219,6 +219,16 @@ class CliParser:
                 'Default %(default)s'
             ),
         )
+        parser.add_argument(
+            '--feed-updater',
+            default="openvas",
+            choices=['openvas', 'nasl-cli'],
+            help=(
+                'Sets the method of updating the feed.'
+                ' Can either be openvas or nasl-cli.'
+                ' Default: %(default)s.'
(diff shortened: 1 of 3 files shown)


ospd/parser.py
```diff
@@ -219,9 +219,27 @@ def __init__(self, description: str) -> None:
...
+            help=(
+                'Sets the method of updating the feed.'
+                ' Can either be openvas or nasl-cli.'
+                ' Default: %(default)s.'
+            ),
+        )
 
         self.parser = parser
 
+    def feed_updater(self, string: str) -> str:
```
bjoernricks, line 235: The function isn't used. I guess you want to use it as the type of the `--feed-updater` argument but instead it would be better to just use a [choice in this case](https://docs.python.org/3/library/argparse.html#choices).

nichtsfrei: Thanks, your right, I forgot to add it as type but choices are better.

jjnicola (approved):

PR #908, opened 2023-05-30
Create VT XML at startup to speed up GET_VTS

mattmundell:
## What

At startup ospd-openvas stores a VT XML string for each VT.  When gvmd calls GET_VTS, then ospd-openvas replies with the premade strings.

## Why

With recent PRs gvmd takes about 2m30s to do a rebuild.  1m30s of that is waiting for ospd-openvas to send the VTs.  With this change gvmd only waits 1s.

## References

Recent gvmd PRs: greenbone/gvmd/pull/1959, greenbone/gvmd/pull/1988

GEA-49

### DIFF
--- a/ospd/command/command.py
+++ b/ospd/command/command.py
@@ -420,8 +420,23 @@ class GetVts(BaseCommand):
 
         yield begin_vts_tag
         if not version_only:
-            for vt in self._daemon.get_vt_iterator(vts_selection, vt_details):
-                yield xml_helper.add_element(self._daemon.get_vt_xml(vt))
+            if vt_details:
+                if vt_filter:
+                    logger.debug('Getting filtered with details.')
+                    for vt_id in self._daemon.get_vt_id_iterator(vts_selection):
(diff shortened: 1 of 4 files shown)


ArnoStiefvater:
Hey @mattmundell 
Thanks a lot for the WIP PR and nice to see you again!

The PR description sounds great. I will nonetheless add a "Request changes" GitHub review just to make sure that it is not accidentally merged at some point without a scanner-dev having a look. Maybe thats not the best workflow but for now it works ¯\\_(ツ)_/¯

ArnoStiefvater (changes requested):
As explained in my other comment. This comment is just here to make sure that a scanner-dev looks at this at some point too.

mattmundell:
Hey @ArnoStiefvater, thanks, nice to see you too :)
I expect it can be done a better way -- mostly I was looking to prove that it can be fast.  Happy to clean it up in whatever way, if devs have time for it.
The VT XML 'cache' uses memory and adds to the startup time.
I still have to test the cases where GET_VTS is given a filter or a VT id.

PR #916, opened 2023-08-02
Fix:  examples/wmiexec.py [Errno 2] No such file or directory

nichtsfrei:
The debian package installed on openvas image is producing

```
 ‘/usr/share/doc/python3-impacket/examples/wmiexec.py’: [Errno 2] No such file or directory
```

To mitigate it until either `python3-impacket` is fixed or openvas does
verify if either `wmiexec.py` or `impacket-wmiexec` is in PATH we
install impacket via pip and link `impacket-wmiexec` to `wmiexec.py`.

### DIFF
--- a/.docker/prod.Dockerfile
+++ b/.docker/prod.Dockerfile
@@ -49,6 +49,9 @@ RUN apt-get update && \
     apt-get remove --purge --auto-remove -y && \
     rm -rf /var/lib/apt/lists/*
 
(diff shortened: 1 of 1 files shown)


ArnoStiefvater (approved):

bjoernricks:
Just for curiosit,y afaik that's required only for windows, so why not add it to openvas-smb instead of ospd-openvas?

PR #949, opened 2023-11-16
Add: check if notus via openvasd is enabled in the openvas config

jjnicola:
## What
Since a new notus backend (rs-notus via openvasd) was integrated with openvas, check if the new backend is enabled and then, do not initialized the mqtt client and daemon for the python notus backend.
Notus result are stored directly in redis kb by openvas.

Jira: SC-947

## Why
A new backend is provided. Mqtt is not needed anymore.

## Checklist

- [ ] Tests


jjnicola:
greenbone/openvas-scanner#1519

Kraemii (dismissed):

PR #1024, opened 2025-02-03
Adds krb5 credential example into OSP.xml

nichtsfrei:
SC-1230


docs/OSP.xml
```diff
@@ -182,6 +192,17 @@ SPDX-License-Identifier: GPL-2.0-or-later
         </credential>
       </e>
     </example>
+    <example>
+      <summary>KRB5 Credentials. The kdc value is only used when no matching realm was found.</summary>
```
ArnoStiefvater, line 196:
```suggestion
      <summary>Kerberos Credentials. The kdc value is only used when no matching realm was found.</summary>
```

PR #1026, opened 2025-02-12
Add a delay after daemon.run() to avoid test failure

MocioF:
## What
Running tests sometimes the test tests.messaging.test_mqtt.MQTTDaemonTestCase fails.
You can see the report here: https://bugs.gentoo.org/934153

## Why
Adding a small delay, seems to avoid the failure so that client.connect.assert_called_with() and client.loop_start.assert_called_with() run after daemon is setup.

The 3 arg on sleep() is arbitrary but in my test it is enough; a loop check to wait until daemon is started would be a better solution.

### DIFF
--- a/tests/messaging/test_mqtt.py
+++ b/tests/messaging/test_mqtt.py
@@ -3,6 +3,7 @@
 #
 # SPDX-License-Identifier: AGPL-3.0-or-later
 
+import time
 from datetime import datetime
 from uuid import UUID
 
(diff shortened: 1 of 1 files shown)


ArnoStiefvater:
Hey @MocioF 

Thanks for creating the PR. We would like to fix the problem but don't think putting a sleep is the best option. We can investigate. If there is time pressure on your side you might want to investigate yourself, we are open for PRs (or changes to this one) as usual.

